PR #10, opened 2020-09-26
rename weak syms and finish splitting files

DarkRTA:


DarkRTA:
dont merge this until i rename the WEAK symbols and split the remaining files

DarkRTA:
this should be good to merge now

PR #19, opened 2020-09-30
decomp xEnv.cpp

mattbruv:


src/Core/p2/iEnv.h
```diff
@@ -1,6 +1,27 @@
 #ifndef IENV_H
 #define IENV_H
 
+extern struct RpWorld;
+extern struct xJSPHeader;
+extern struct RpLight;
+extern struct RwFrame;
```
DarkRTA: Opaque struct definitions don't need `extern` unless there's some weird mwcceppc quirk I'm unaware of.

They also don't belong here in the first place but should be in the correct header file.

src/Core/p2/iEnv.h
```diff
@@ -1,6 +1,27 @@
...
+	xJSPHeader* jsp;
+	RpLight* light[2];
+	RwFrame* light_frame[2];
+	int memlvl;
+};
+
+void iEnvFree(iEnv* param_1);
+void iEnvRender(iEnv* param_1);
+void iEnvDefaultLighting(iEnv* param_1);
+//void iEnvLoad(iEnv* env, void* data, int dataType);
```
DarkRTA: Is there a reason why this is commented out?

src/Core/x/xEnv.cpp
```diff
@@ -1 +1,38 @@
-#include "xEnv.h"
\ No newline at end of file
+#include "xEnv.h"
+
+extern xEnv* gCurXEnv;
```
DarkRTA, line 4: I'd leave a todo comment here saying this needs to be moved from xEntMotion.s

src/Core/x/xEnv.h
```diff
@@ -1,7 +1,18 @@
 #ifndef XENV_H
 #define XENV_H
 
-typedef struct xEnv;
+#include "../../dolphin/include/dolphin/types.h"
+#include "../p2/iEnv.h"
+
+extern struct xLightKit;
```
DarkRTA: Same deal with the first group of opaque structs. This doesn't need extern and should be defined in xLightKit.h.

DarkRTA:
Also you should be using the int types like `s32` from `<dolphin/types.h>` (btw we really need to fix that include path as it doesn't work at the moment)

seilc:
We should not be using the types from the dolphin types header. Those are reserved for the Dolphin SDK only

seilc:
We shouldn't even need to include dolphin headers unless we specifically need a struct or something from them

DarkRTA:
> We should not be using the types from the dolphin types header. Those are reserved for the Dolphin SDK only

This is about the integer types with a known width. If using `dolphin/types.h` is not an option we should clone what we need from it and use our own `types.h`.

`int` can have a different meaning depending on what you compile for whereas something like `s32` is known to be a 32 bit signed int if the typedef is correct.

And the assumption that we are currently targeting GCN is only true at the moment. Later on down the line this codebase is *going* to be ported to other platforms at some point, and that port will have a lot of problems if we aren't using known width integer types.

DarkRTA:
Also, if we aren't supposed to be using the Dolphin SDK `types.h` outside of the Dolphin SDK, then we should *really* fix this include:

https://github.com/pslehisl/bfbbdecomp/blob/b2ebb268b29038e41c4ad2a5ccd12f3507c39d36/src/Game/zParEmitter.cpp#L3

DarkRTA:
Oh yeah i'd also remove xEnv.s if we no longer need it.

mattbruv:
I fully decomped another file just now. What do we need to do to get this merged?

DarkRTA:
Address the issues i pointed out earlier in the thread

PR #22, opened 2020-10-06
Setup zEntPLayer.cpp and decompile some functions

mattbruv:


include/rwsdk/rwcore.h
```diff
@@ -72,6 +72,25 @@ struct RwTexture
...
+{
+	RwV3d right;
+	unsigned int flags;
+	RwV3d up;
+	unsigned int pad1;
+	RwV3d at;
+	unsigned int pad2;
+	RwV3d pos;
+	unsigned int pad3;
+};
```
seilc: These structs should use the `RwReal` and `RwUInt32` types (defined at the top of the file) instead of `float` and `unsigned int`.

In the future I would recommend referencing the official RW SDK when copying types from the PS2 data:
* https://github.com/resigma/rwsdk-v35-ps2
* https://github.com/resigma/rwsrc-v37-pc

src/Core/x/xModel.h
```diff
@@ -1,10 +1,67 @@
...
+#include "xSurface.h"
 #include "xMath3.h"
 
+struct xModelInstance;
+
+// TODO: this probably can be simplified
+struct _class
+{
+	xVec3* verts;
+};
```
seilc: The name `_class` is usually generated for unnamed structs. I would define this struct as part of the declaration for `xModelInstance::anim_coll`, like this.

```
struct xModelInstance
{
    /* ... */
    struct
    {
        xVec3* verts;
    } anim_coll;
};
```

seilc (changes requested):

PR #26, opened 2020-10-23
Account for all functions in symbol file

mattbruv:
All the functions in the symbol file for our relevant game sources are now included in pragmas.

I also changed it so that above every function, there is a comment like this:

```
// func_8010F1EC
#pragma GLOBAL_ASM("asm/Core/x/xRMemData.s", "__dl__10RyzMemDataFPv"
```

That way, a search and replace will be trivial when we decide on the proper way to document the function. I suppose down the road we'd want to substitute in the right typedefs, etc. but I made it this way to allow it to be easy in the future while not having to wait on it for now.

Also there were changes to the pragma processor because there was a bug with it.


mattbruv:
Also you definitely want to squash all these commits into a single one if possible

PR #29, opened 2020-10-24
make the build process "quieter"

DarkRTA:
This commit modifies the Makefile in order to allow for a quieter build. It does hide the executed commands but they can be shown by running `make S=`. This also fixes the build on my machine by commenting out a line that apparently does nothing useful. I left it in on the rare chance that we do need it.


DarkRTA:
mp told me that he accidentally committed that line

gonna go ahead and outright remove it

PR #30, opened 2020-10-25
partially match iEnv

DarkRTA:
sorry about not naming shit properly but the more important part is that this still OKs

i do have one thing i'm extremely unsure about tho

### DIFF
--- a/src/Core/p2/iEnv.h
+++ b/src/Core/p2/iEnv.h
@@ -1,6 +1,5 @@
 #ifndef IENV_H
 #define IENV_H
-
 #include <rpworld.h>
 #include <rwcore.h>
 #include "../x/xJSP.h"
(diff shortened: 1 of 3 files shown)


src/Core/p2/iEnv.h
```diff
@@ -17,9 +16,15 @@ struct iEnv
     int32 memlvl;
 };
 
+// bringing in xEnv.h breaks the build entirely
+struct xEnvAsset;
```
DarkRTA, line 20: Alright so I am very unsure about this.

What do you want me to do here?

seilc: xEnv.h includes iEnv.h already - it has to do this because the xEnv struct in xEnv.h uses the complete iEnv type - so including iEnv.h in xEnv.h creates a circular dependency which breaks the build.

What you did here is correct, forward declaring xEnvAsset in iEnv.h since it is only used as a pointer in iEnv.h.

PR #35, opened 2020-11-03
Decomp Part Of zMusic.cpp

Gota7:
Progress.


src/Game/zMusic.h
```diff
@@ -1,6 +1,36 @@
...
+	uint32 loop;
+	int32 unk;
+	uint32 assetID;
+	float32 lastVol;
+};
+
+enum LevelMusic {
+	//TODO: POPULATE MUSIC LISTING!
+	Null
+};
```
mattbruv: It appears that this enum doesn't exist in the dwarf data and that this was made up. Avoid making up enums or types that don't exist anywhere in the mangled names or dwarf data. We want to keep the codebase strictly to how they typed things for now. Documentation or cleaning up things like this will be done at some point far in the future (if at all).

You're using your custom enum here, in a commented out function:
`LevelMusic getCurrLevelMusicEnum()`

The type of this function in the dwarf data is an `int32`: 
`int32 getCurrLevelMusicEnum();`

Was it an enum in the original source? Probably. Maybe there is an enum for this somewhere in the dwarf data, but if you can't find it, just leave it as an int. This also means you shouldn't be casting every single int into your new enum type.

src/Game/zMusic.cpp
```diff
@@ -29,17 +121,91 @@
...
 // func_800A7C60
 #pragma GLOBAL_ASM("asm/Game/zMusic.s", "zMusicSetVolume__Fff")
 
-// func_800A7C90
-#pragma GLOBAL_ASM("asm/Game/zMusic.s", "zMusicReset__Fv")
+//Not sure what's wrong with this one. Doesn't match in the slightest.
+/*void zMusicSetVolume(float32 vol, float32 delay)
+{
+	if (delay <= minDelay) {
+    	volume.end = vol;
```
mattbruv: Please try to always format every single `cpp` or `h` file that you've changed with `clang-format`. We have a [formatting style](https://github.com/pslehisl/bfbbdecomp/blob/master/.clang-format) which we are trying to stick to, to keep all code consistent.

Note that `clang-format` won't fix formatting on anything inside a comment, so you'll have to uncomment it, format it, and then re-comment it.

src/Game/zMusic.cpp
```diff
@@ -1,19 +1,111 @@
...
+  	}
+}
 
 // func_800A6F50
 #pragma GLOBAL_ASM("asm/Game/zMusic.s", "zMusicInit__Fv")
 
 // func_800A7314
 #pragma GLOBAL_ASM("asm/Game/zMusic.s", "getCurrLevelMusicEnum__Fv")
 
+//Correct, but won't work due to the switch case jump table (messes with offsets).
```
mattbruv: Instead of commenting out the function, I think this is slightly better:
```cpp
#if 0
// put entire function here as if it weren't commented out.
// also add any comments related to the non-matching status above the function here, but within the #if scope.
#else
// func_800A7414
#pragma GLOBAL_ASM("asm/Game/zMusic.s", "zMusicDo__Fi")
#endif
```
That way, anyone who comes along can simply change the `0` to `1` and mess with it without dealing with comments. This is kind of nitpicky, probably not that big of a deal, and we haven't come to a group decision about how to include non-matching functions in the source code, but I prefer it.

(After reading other formatting comment): Note that clang-format won't format code inside an unreachable directive, so before you set it to `#if 0`, format the code inside it first.

mattbruv (changes requested):

PR #36, opened 2020-11-03
Decomp Part of zMusic.cpp

Gota7:
Formatting fixes.


src/Game/zMusic.cpp
```diff
@@ -7,104 +7,111 @@
...
-  			xSndSetVol(track->snd_id, volume.cur); //Only reset volume if the sound ID is not blank.
-  		}
-  	}
+    for (int32 i = 0; i < 2; i++)
+    {
+        zMusicTrackInfo* track = &sMusicTrack[i];
+        if (track->snd_id != 0)
+        {
+            xSndSetVol(track->snd_id,
+                       volume.cur); // Only reset volume if the sound ID is not blank.
```
mattbruv: Nitpicking, but I'd put this comment above the function call instead of after to keep the function call on a single line

src/Game/zMusic.h
```diff
@@ -3,30 +3,27 @@
...
-};
-
-enum LevelMusic {
-	//TODO: POPULATE MUSIC LISTING!
-	Null
+struct zMusicTrackInfo
+{
+    uint32 snd_id;
+    uint32 loop;
+    int32 situation; // Is actually a zMusicSituation*.
```
mattbruv: I know it's a pointer and 4 bytes anyway, but if copying dwarf struct definitions into the source code, please copy them as they are without changing any types. Sometimes this is annoying because you have to go down a rabbit hole of defining structs, but it will be helpful to others down the road if we import everything as we come across it.

https://github.com/pslehisl/bfbbdecomp/blob/7437353b1ae66044e80bed5c02704a234ca25609/dwarf/SB/Game/zMusic.cpp#L1949-L1962

This struct is easy to import.

mattbruv (changes requested):
Thanks for the fixes. Just a few small things and I'll merge your PR.

.clang-format
```diff
@@ -20,7 +20,7 @@ BinPackArguments: true
 BinPackParameters: true
 BraceWrapping:
   AfterClass: true
-  AfterCaseLabel: true
+  #AfterCaseLabel: true
```
mattbruv: Why'd you comment this out?

src/Game/zTaxi.h
```diff
@@ -1,6 +1,67 @@
...
+        ANSWER_YES,
+        ANSWER_NO,
+        ANSWER_NONE
+    };
+};
+
+struct taxiCB : callback
+{
+    //zTaxi* taxi;
+    //answer_enum answer;
```
mattbruv: Similar to my other comment, please import structs exactly how they appear in the dwarf data (ONLY unless there is a reason to delete a struct member, i.e. the dwarf data is incorrect between PS2 and gamecube).

Gota7: That, I commented out because I couldn't understand why it didn't compile, only to see I forgot ztalkbox:: in front of the answer enum.

The answer enum is using 1 and not both, as defining both leads to errors in compilation and makes no sense as they are nearly the same.

.clang-format
Gota7: I need to comment it out in order for my clang-format to work and I forgot to exclude changes.

PR #38, opened 2020-11-23
Fix python tooling bits

stravant:
* Invoke python via "py -3" command instead of "python" to avoid python
  version issues for systems with multiple python versions installed.

* Add .pyc to the gitignore so that cached python files aren't picked
  up by git.


mattbruv:
I get the following error trying to build with powershell using this new makefile: `make: py: Command not found`. Is this way of invoking python common? Can you give some more info on this?

stravant:
That's weird, I thought that every Python 3 install comes with the py command. I don't normally work with Python toolchains, I will try to find what the right way to do this is.

DarkRTA:
using `python3` to invoke python 3.x should also be tested

i've never heard of `py` until now

mattbruv:
Yeah, I didn't know if it was just me but I've also never heard of/used `py` in my life

stravant:
Yeah, I don't see a reason not to use `python3` as the command, changed it. I just went to `py -3` because I've had to edit build scripts before where I saw that approach used. They must have been set up to make sure that that command was included in the path somehow.

stravant:
@mattbruv Could you merge this now?

PR #39, opened 2020-11-27
Add "inspect" make target

stravant:
* Add an "inspect" make target which builds a file specified with the
  INSPECT make variable to inspect.s in the root directory to experiment
  with a function being reverse engineered.

* Also turn off warnings on the linker so that you don't get warning
  spam when running the make command.


stravant:
Will include the symbol ripper and better inspect capability etc in a different PR.

PR #40, opened 2020-11-27
zCombo.cpp Progress

stravant:
* 100% match for: fillCombo, zComboHideMessage, and ztextbox::visible.

* 100% match but suspicious for zCombo_HideImmediately.

* Close match minus instruction ordering for zCombo_Add, zCombo_Setup,
  put the current close matches behind #if 0.

* Identified most of the names for most of the variables the file uses.

* Filled out needed stuff in xHud header files.

### DIFF
--- a/asm/Game/zCombo.s
+++ b/asm/Game/zCombo.s
@@ -77,17 +77,17 @@ lbl_801959F0:
 zCombo_Setup__Fv:
 /* 80195A00 00192800  94 21 FF D0 */	stwu r1, -0x30(r1)
 /* 80195A04 00192804  7C 08 02 A6 */	mflr r0
-/* 80195A08 00192808  3C 60 80 27 */	lis r3, lbl_80274A00@ha
+/* 80195A08 00192808  3C 60 80 27 */	lis r3, zCombo_Strings@ha
 /* 80195A0C 0019280C  90 01 00 34 */	stw r0, 0x34(r1)
-/* 80195A10 00192810  38 63 4A 00 */	addi r3, r3, lbl_80274A00@l
+/* 80195A10 00192810  38 63 4A 00 */	addi r3, r3, zCombo_Strings@l
 /* 80195A14 00192814  38 00 00 00 */	li r0, 0
 /* 80195A18 00192818  BE E1 00 0C */	stmw r23, 0xc(r1)
 /* 80195A1C 0019281C  38 63 00 C1 */	addi r3, r3, 0xc1
-/* 80195A20 00192820  90 0D 98 E0 */	stw r0, lbl_803CC1E0-_SDA_BASE_(r13)
-/* 80195A24 00192824  90 0D 98 DC */	stw r0, lbl_803CC1DC-_SDA_BASE_(r13)
-/* 80195A28 00192828  90 0D 98 D8 */	stw r0, lbl_803CC1D8-_SDA_BASE_(r13)
-/* 80195A2C 0019282C  C0 02 B2 60 */	lfs f0, lbl_803CFBE0-_SDA2_BASE_(r2)
-/* 80195A30 00192830  D0 0D 98 E4 */	stfs f0, lbl_803CC1E4-_SDA_BASE_(r13)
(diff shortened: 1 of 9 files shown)


asm/Game/zCombo.s
```diff
@@ -77,17 +77,17 @@ lbl_801959F0:
...
 /* 80195A1C 0019281C  38 63 00 C1 */	addi r3, r3, 0xc1
-/* 80195A20 00192820  90 0D 98 E0 */	stw r0, lbl_803CC1E0-_SDA_BASE_(r13)
-/* 80195A24 00192824  90 0D 98 DC */	stw r0, lbl_803CC1DC-_SDA_BASE_(r13)
-/* 80195A28 00192828  90 0D 98 D8 */	stw r0, lbl_803CC1D8-_SDA_BASE_(r13)
-/* 80195A2C 0019282C  C0 02 B2 60 */	lfs f0, lbl_803CFBE0-_SDA2_BASE_(r2)
-/* 80195A30 00192830  D0 0D 98 E4 */	stfs f0, lbl_803CC1E4-_SDA_BASE_(r13)
+/* 80195A20 00192820  90 0D 98 E0 */	stw r0, comboCounter-_SDA_BASE_(r13)
+/* 80195A24 00192824  90 0D 98 DC */	stw r0, comboLastCounter-_SDA_BASE_(r13)
+/* 80195A28 00192828  90 0D 98 D8 */	stw r0, comboPending-_SDA_BASE_(r13)
+/* 80195A2C 0019282C  C0 02 B2 60 */	lfs f0, someComboFloatConstant-_SDA2_BASE_(r2)
```
mattbruv: For now, please only name labels the exact name they have in the symbols.txt file (if they have one). Renaming things can come later, but we don't want someone to rename something innacurately and have it confuse someone down the line. It's better to leave unnamed labels `lbl_whatever` so at least we know that we don't know what they are. Please revert any of these custom names to their original names or back to their original unnamed form.

src/Core/x/xHud.h
```diff
@@ -1,6 +1,83 @@
...
+	motive_node* _motive_top;
+	motive_node* _motive_temp;
+	motive_node** _motive_temp_tail;
+
+	widget(asset& a);
+	void setup();
+    void enable();
+    void update(float32 dt);
+	void dispatch(uint32 event);
+	void clear_motives(uint8(*fp_update)(widget&, motive&, float32), void* context);
```
mattbruv: Please run `clang-format` on all of this code so it has a consistent style.

src/Game/zCombo.cpp
```diff
@@ -2,23 +2,208 @@
...
+{
+    w.hide();
+}
 
 // func_80195C8C
-#pragma GLOBAL_ASM("asm/Game/zCombo.s", "zCombo_HideImmediately__Fv")
+void zCombo_HideImmediately()
+{
+    if (comboHUD != NULL) {
+        // This is probably wrong, but generates the right code
```
mattbruv: I would add a `// TODO: blah blah blah` comment here, so anyone who searches for TODO can find it.

mattbruv (changes requested):

src/Core/x/xHud.h
stravant: I made this PR before I realized that the `dwarf/SB` files I was copying from used tabs while our code uses spaces, I've got this down now for the subsequent code.

asm/Game/zCombo.s
stravant: Like, _all_ of them?

Some of them such as `extern zComboReward comboReward[16];` are completely unambiguous: In assuming the type that the variable has there is already only one possible name for it in the `dwarf/SB` symbols. Presumably renaming them is sensible?

Same thing for floating point constants. They get referenced in a bunch of places in the file, so if I do work to find out what value the floating point constant has it makes sense to rename the symbol to the value that the floating point constant has so that someone in the future doesn't have to do that work again.

I reverted the floating point / integer variables which aren't 100% certain to ambiguous names.

mattbruv: @DarkRTA @pslehisl what do you guys think

PR #52, opened 2020-12-08
Decompiled 2 functions in iMath3.cpp as well as partaily iSphereIsectRay

sonich2401:
Decompiled 2 Functions as well as partailly iSphereIsectRay. Also in progress to decompile Vec3Copy

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,12 +22,5 @@ tools/elf2dol
 tools/lmgr8c.dll
 tools/lmgr326b.dll
 ldscript.lcf
-asmdiff.sh
 .clang-format
 SpongeBob SquarePants - Battle for Bikini Bottom.7z
-ldscript.lcf
-asmdiff.sh
-.clang-format
-ldscript.lcf
-asmdiff.sh
-.clang-format
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,15 @@ tools/elf2dol
 *.cp
 .vscode
 .pragma
+tools/lmgr8c.dll
+tools/lmgr326b.dll
+ldscript.lcf
+asmdiff.sh
+.clang-format
+SpongeBob SquarePants - Battle for Bikini Bottom.7z
+ldscript.lcf
+asmdiff.sh
+.clang-format
+ldscript.lcf
+asmdiff.sh
+.clang-format
--- a/src/Core/p2/iMath3.cpp
+++ b/src/Core/p2/iMath3.cpp
@@ -1,6 +1,4 @@
 #include "iMath3.h"
-#include "../x/xVec3Inlines.h"
-#include <types.h>
 
 // func_800C4F18
 void iMath3Init(){
--- a/src/Core/p2/iMath3.cpp
+++ b/src/Core/p2/iMath3.cpp
@@ -1,5 +1,5 @@
 #include "iMath3.h"
-
+#include <types.h>
 // func_800C4F18
 void iMath3Init(){
 
--- a/src/Core/p2/iMath3.cpp
+++ b/src/Core/p2/iMath3.cpp
@@ -1,5 +1,7 @@
 #include "iMath3.h"
+
 #include <types.h>
+
 // func_800C4F18
 void iMath3Init(){
 
--- a/src/Core/p2/iMath3.cpp
+++ b/src/Core/p2/iMath3.cpp
@@ -1,16 +1,115 @@
 #include "iMath3.h"
-
+#include "../x/xVec3Inlines.h"
 #include <types.h>
 
 // func_800C4F18
-#pragma GLOBAL_ASM("asm/Core/p2/iMath3.s", "iMath3Init__Fv")
+void iMath3Init(){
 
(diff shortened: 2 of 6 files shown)


src/Core/p2/iMath3.cpp
```diff
@@ -1,16 +1,115 @@
 #include "iMath3.h"
-
+#include "../x/xVec3Inlines.h"
 #include <types.h>
 
 // func_800C4F18
-#pragma GLOBAL_ASM("asm/Core/p2/iMath3.s", "iMath3Init__Fv")
+void iMath3Init(){
 
+}
```
mattbruv, line 8: Make sure you format every file you commit with clang-format with our ruleset

.gitignore
```diff
@@ -19,3 +19,15 @@ tools/elf2dol
...
+ldscript.lcf
+asmdiff.sh
+.clang-format
+SpongeBob SquarePants - Battle for Bikini Bottom.7z
+ldscript.lcf
+asmdiff.sh
+.clang-format
+ldscript.lcf
+asmdiff.sh
+.clang-format
```
mattbruv: I don't know if a program generated this stuff or if you handwrote it, but none of this should be here. A lot of those things are scripts that we do want tracked by git if they are changed.

src/Core/p2/iMath3.cpp
```diff
@@ -1,16 +1,115 @@
...
 // func_800C4F1C
-#pragma GLOBAL_ASM("asm/Core/p2/iMath3.s", "iSphereIsectVec__FPC7xSpherePC5xVec3P6xIsect")
-
+void iSphereIsectVec(const xSphere *param_1,const xVec3 *param_2,xIsect *param_3)
+{
+  float fVar1;
+  xVec3Sub(&param_3->norm,param_2,&param_1->center);
+  fVar1 = xVec3Length(&param_3->norm);
+  param_3->dist = (float)fVar1;
```
mattbruv, line 15: use our custom typedefs: i.e. `float32` instead of `float`

src/Core/p2/iMath3.cpp
```diff
@@ -1,16 +1,115 @@
...
+        }
+      }
+    }
+  }
+  __psq_l0(auStack8,in_GQR0);
+  __psq_l1(auStack8,in_GQR0);
+  __psq_l0(auStack24,in_GQR0);
+  __psq_l1(auStack24,in_GQR0);
+  return;
+  #endif
```
mattbruv, line 110: I get that this  probably is close to matching, but it's pretty useless as far as readable code goes. If you compare this function to `iSphereIsectVec`, it is much more legible and correct than this. This looks like you basically copied and pasted ghidra's output into a comment, which doesn't really help much honestly. If it almost matches and you want to keep it, this function needs to be cleaned up A LOT. Use the struct definitions to set the properties of a type, not random pointer math everywhere. Basically make this look like `iSphereIsectVec`, and use our custom types everywhere, and format the code with clang-format before I can accept this. Alternatively, you can submit the two other functions in a separate PR and I can merge them (after other issues are addressed), but I don't want to merge this code in

src/Core/p2/iMath3.h
```diff
@@ -1,6 +1,42 @@
...
+	xVec3 dir;
+	float32 min_t;
+	float32 max_t;
+	int flags;
+};
 
+
+#define float32 float
+#define uint32 unsigned int
+
```
mattbruv: You should include `<types.h>` instead of re-defining our types

src/Core/x/xVec3.cpp
```diff
@@ -9,7 +9,12 @@
...
 // func_8004E4DC
-#pragma GLOBAL_ASM("asm/Core/x/xVec3.s", "xVec3Copy__FP5xVec3PC5xVec3")
-
+//#pragma GLOBAL_ASM("asm/Core/x/xVec3.s", "xVec3Copy__FP5xVec3PC5xVec3")
+void xVec3Copy(xVec3 * o, const xVec3 * a){
+    float tmp1[] = {((float*)(void*)a)[0],((float*)(void*)a)[1]};
+    float tmp2[] = {((float*)(void*)a)[2],((float*)(void*)a)[3]};
+    *o = *(xVec3*)tmp1;
+    *(float*)(o+0x8)= tmp2[0];
+}
```
mattbruv, line 18: This doesn't 🆗 right? I can't merge this code if it's uncommented and doesn't build properly. Also, as discussed in discord, this function was probably hand-written assembly, so I highly doubt that's anything close to what they wrote

mattbruv (changes requested):

src/Core/p2/iMath3.cpp
sonich2401: Yea I was on 0 sleep when I made this PR. Tbh i shouldn't have made this PR yet. How about I just delete this PR until I have my bearings straight?

sonich2401:
Closed Until I can get my act together and learn how to make a proper PR lol.

PR #53, opened 2020-12-19
xMemMgr.cpp 5% progress

stravant:
* Hard-fought 100% match for xMemInitHeap.

* Attempted several other functions but all the functions in this file
  are proving extremely difficult to 100% match thanks to using a ton
  of temporary variables in their bodies on top of the underlying code
  being complicated to begin with.


DarkRTA:
lgtm but someone else should look at this

stravant:
Do you think it's worth merging the three attempted but not close functions at all?

DarkRTA:
>Do you think it's worth merging the three attempted but not close functions at all?

absolutely

PR #62, opened 2020-12-24
iAnimSKB.cpp 5% complete

seilc:
* Matching: iAnimDurationSKB
* Non-matching: iAnimEvalSKB
* Not attempted:
  * _iAnimSKBAdjustTranslate, _iAnimSKBExtractTranslate
  * std::fabsf, fabs - Not sure what to do with these two since nothing in iAnim.cpp seems to reference them.

### DIFF
--- a/src/Core/p2/iAnimSKB.cpp
+++ b/src/Core/p2/iAnimSKB.cpp
@@ -5,8 +5,10 @@
 // func_800BF1C0
 #pragma GLOBAL_ASM("asm/Core/p2/iAnimSKB.s", "iAnimEvalSKB__FP14iAnimSKBHeaderfUiP5xVec3P5xQuat")
 
-// func_800BF7B0
-#pragma GLOBAL_ASM("asm/Core/p2/iAnimSKB.s", "iAnimDurationSKB__FP14iAnimSKBHeader")
+float32 iAnimDurationSKB(iAnimSKBHeader* data)
+{
+    return ((float32*)((iAnimSKBKey*)(data + 1) + data->KeyCount))[data->TimeCount - 1];
+}
 
 // func_800BF7D0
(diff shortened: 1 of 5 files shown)


src/Core/p2/iAnimSKB.cpp
```diff
@@ -1,12 +1,140 @@
...
+            tran->z =
+                lerp * (scalez * k[1].Tran[2] - scalez * k[0].Tran[2]) + scalez * k[1].Tran[2];
+        }
+    }
+}
+#endif
 
-// func_800BF7B0
-#pragma GLOBAL_ASM("asm/Core/p2/iAnimSKB.s", "iAnimDurationSKB__FP14iAnimSKBHeader")
+float32 iAnimDurationSKB(iAnimSKBHeader* data)
```
mattbruv, line 134: Looks like you forgot to include <types.h> in the CPP file, I'm getting errors in vscode at this function, although the PR builds and is :OK:

seilc: Weird, I'm not getting any errors in vscode at this function. <types.h> does get included indirectly (iAnimSKB.h > xMath3.h > xVec3.h > types.h), so it shouldn't need to be directly included in iAnimSKB.cpp.

mattbruv: Do you have the C/C++ extension installed? That's the one that gives me an error

seilc: Yeah, it's able to find the types just fine on my end. I have these include paths and defines set up as well:

```
${workspaceFolder}/**
${workspaceFolder}/include/**
```

```
_DEBUG
UNICODE
_UNICODE
GAMECUBE
```

stravant (approved):
I like to add a `#include <types.h>` at the start of every header file I work on, though it's not required.

🆗

PR #63, opened 2020-12-24
zRenderState 100% complete

stravant:
* We could translate the constants in the giant block of state setting
  to the equivalent enums too, but I'm not confident enough that I would
  get that 100% correct because there's no typesafey to ensure we use
  the correct kind of enum for each call.


mattbruv (approved):
Good job 🆗

PR #64, opened 2020-12-25
iEnv.cpp 90% complete

seilc:
Progress from where Dark left off, cleaned up the existing functions and fully decomped the rest. Almost all functions are matching, only iEnvLoad is slightly non-matching.


mattbruv (approved):
🆗

PR #68, opened 2020-12-26
xutil.cpp 80% complete

seilc:
All functions matching except for xUtil_crc_update and xUtil_wtadjust.

Note: I used `char` instead of `int8` in places that involve strings and data buffers. This is intentional, I think `char` is a much more appropriate type than `int8` in those types of cases. IMO we should all be doing this wherever we feel necessary, it'll make the code more portable.


mattbruv (approved):
Great work

PR #69, opened 2020-12-26
iLight.cpp 100%-ish complete

seilc:
All functions matching 🎉 however... iLight.s cannot be removed yet because there are 2 float constants that need to stay there in order for iLightCreate to match at the moment.


mattbruv (approved):
great-ish job

PR #70, opened 2020-12-27
iMath3.cpp 90% complete

seilc:
All functions matching except for iBoxIsectRay and iBoxIsectSphere.

Also moved xVec2's definition to xMath2.h; in the dwarf data, xVec2 actually has a few member functions (xVec2::length(), xVec2::normal(), etc.) which are defined in xMath2.h, so that must be where the struct definition for xVec2 is.


mattbruv (approved):
Looks good, I hope it OKs and there's no formatting problems 😅

PR #77, opened 2020-12-29
zActionLine 50%+

mattbruv:
Decompiled two functions, and attempted the other two. They should be close, or at least at a very good starting point.

I also moved stravant's walkthrough and tips file into a new `docs` folder so it doesn't take up space in the root folder


stravant (approved):
:OK:

Don't forget to squash the commits merging this.

PR #79, opened 2021-01-01
zVar 80% progress

stravant:
* Attempt all functions in file, almost all 100% match.

* A couple of larger ones are left NON_MATCHING.

* A couple are doing really weird stuff that I have no clue on.


mattbruv (approved):
🆗

PR #80, opened 2021-01-02
Fix build error

mattbruv:
…ustom types

### DIFF
--- a/include/rwsdk/rwcore.h
+++ b/include/rwsdk/rwcore.h
@@ -1,6 +1,7 @@
 #ifndef RWCORE_H
 #define RWCORE_H
 
+#include <types.h>
 #include "rwplcore.h"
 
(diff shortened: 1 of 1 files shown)


seilc (changes requested):
rwsdk is its own thing and should not be including our types header. _RxObjSpace3DVertex should not use our types, it should be using RenderWare's types:

```c
struct _RxObjSpace3DVertex
{
    RwV3d objVertex;
    RxColorUnion c;
    RwV3d objNormal;
    float32 u;
    float32 v;
};
```

needs to be changed to:

```c
struct _RxObjSpace3DVertex
{
    RwV3d objVertex;
    RxColorUnion c;
    RwV3d objNormal;
    RwReal u;
    RwReal v;
};
```

and the `#include <types.h>` needs to be removed

seilc (approved):

PR #81, opened 2021-01-02
Partial Decompilation Of iTime.cpp

Gota7:
Decompiled most of the functions.


mattbruv (changes requested):
Why did you add back in `asm/Game/zScript.s`? This file is 100% decompiled and shouldn't be added back. Other than this, your PR looks good.

Gota7:
Not sure why it didn't delete that when fetching; should be fixed now, as long as merging with the latest and some progress on behaviour.

mattbruv (approved):

PR #83, opened 2021-01-02
xCamera.cpp 80% complete

seilc:
64/77 functions matching.

Also, I came across an issue where a few inline functions from `<cmath>` don't end up at the right addresses, so for now all inline functions from `<cmath>` are just forward declarations wrapped by an `#ifndef INLINE`, and the inline definitions are in the `#else` part. Since the inline definitions are currently ifdef'd out, I copied all their definitions to the source files that use them, mirroring how they appear in their declarations.

For example, here's the declaration and (disabled) definition of `std::expf` in `<cmath>`:

```c
namespace std
{
#ifndef INLINE
    // ...
    float expf(float x);
    // ...
#else
    // ...
    inline float expf(float x)
    {
        return (float)exp((double)x);
    }
    // ...
#endif
```

Here's the definition that actually gets compiled in `xCamera.cpp`:

```c
#ifndef INLINE
float std::expf(float x)
{
    return (float)exp((double)x);
}
#endif
```

I recommend we do this for any new `<cmath>` functions that we come across, at least until we figure out how inline functions are ordered relative to surrounding functions.


mattbruv (approved):
🆗

PR #96, opened 2021-01-06
Partial Progress On xGroup.cpp

Gota7:
Most of the functions are done or almost done, two of the remaining 3 are very close to being OK.


src/Core/x/xGroup.h
```diff
@@ -7,6 +7,7 @@ struct xGroupAsset : xBaseAsset
 {
     uint16 itemCount;
     uint16 groupFlags;
+    uint32* link;
```
seilc: `link` should not be added here, it's intentional that it's not included in the original struct. Usually the way xLinkAsset* works is it points to the end of the asset struct, like this:

```
(xLinkAsset*)(asset + 1)
```

I see you did it like this in `xGroupInit`:

```
(xLinkAsset*)(&asset->link)
```

which, yes, does technically accomplish the same thing, but it's not how it was originally done. It might also be the reason behind your non-match.

src/Core/x/xGroup.cpp
```diff
@@ -1,39 +1,177 @@
...
+// There is an instruction swap that seems to be throwing everything off.
+void xGroupInit(xBase* b, xGroupAsset* asset)
+{
+    xBaseInit(b, (xBaseAsset*)asset);
+    b->eventFunc = xGroupEventCB;
+    ((xGroup*)b)->asset = asset;
+    if (b->linkCount)
+    {
+        b->link =
+            (xLinkAsset*)(&asset->link + ((xGroup*)b)->asset->itemCount); // lwz and lhz swap here.
```
seilc:
```
(xLinkAsset*)(&asset->link
```

will need to be changed to

```
(xLinkAsset*)(asset + 1
```

in regards to my other comment.

seilc (changes requested):
One thing needs to be changed, see my comments. Everything else looks good though

src/Core/x/xGroup.cpp
seilc: My bad, you'll probably need to cast `asset + 1` to a uint32* actually so it stays correct: 
```
(xLinkAsset*)((uint32*)(asset + 1) + ((xGroup*)b)->asset->itemCount)
```

seilc (approved):
👍

PR #115, opened 2021-01-13
zCamera

peb-adr:
So this is a few small zCamera functions decompiled.

I also have gotten zCameraReset, zCameraSetLongbounce and zCameraSetHighbounce to where they are functionally equal but dont match due to small register usage bs.
Should in include these in the PR too? perhaps in a `#IF 0` block


src/Game/zCamera.cpp
```diff
@@ -2,37 +2,179 @@
...
+extern float32 zcam_far_pitch;
+extern float32 zcam_wall_d;
+extern float32 zcam_wall_h;
+extern float32 zcam_wall_pitch;
+extern float32 zcam_above_d;
+extern float32 zcam_above_h;
+extern float32 zcam_above_pitch;
+extern float32 zcam_below_d;
+extern float32 zcam_below_h;
+extern float32 zcam_below_pitch;
```
mattbruv, line 44: The naming convention we've generally been giving floats is:

[name of file]\_f\_[float value], so for example:

zCamera_f_1_25 or zCamera_float_1_25  for float 1.25 in zCamera. `below_h` doesn't really tell us anything useful, and since the same float value can be re-used in multiple places in code, the names you give them may not make sense when the float is used in other areas. For that reason, we want to keep them named something general.

mattbruv:
1. Your PR doesn't 🆗 
2. Yes, please include work wrapped in an #ifdef if you've started on/or gotten a function close to matching.

peb-adr:
yeah i see it doesnt OK, im currently working out where it went wrong...

src/Game/zCamera.cpp
peb-adr: o i just pasted them from the dwarf data, assuming those were the names the devs originally chose.
These are not constants btw, so think they should stay like this (?)

mattbruv: Oh, my bad, I thought they were custom names that you chose for the float constants that the compiler generated. If they are variables that the devs created, leave them named like that.

peb-adr:
this should fix the issues, @mattbruv can you reinvoke the checks?

peb-adr:
nvm it did automatically 👍

src/Game/zCamera.cpp
```diff
@@ -2,37 +2,179 @@
...
+extern float32 rewardTiltTime;
+extern float32 rewardTiltAmount;
+// extern zGlobals globals;
+extern xVec3 g_O3;
+extern float32 gSkipTimeFlythrough;
+
+extern float32 lbl_803CD1A8;
+extern float32 lbl_803CD1AC;
+extern float32 lbl_803CD1B0;
+extern float32 lbl_803CD1B4;
```
peb-adr: these however are constants. But i have no idea what their values are. Is there a way to find out other than estimating?

mattbruv: Go to where they are defined in the assembly. It will have an incbin of 4 bytes. Go to the location in the baserom where it is importing the 4 bytes those 4 bytes are the float value, use a tool in a hex editor or online to get the value of those hex bytes to a 32 bit float

mattbruv: If you want to rename them now, you can, and I'll wait to merge the PR. If you don't want to do it, let me know and I'll merge it now

peb-adr: No please hold on with the merge. Ill do it and i also will include the non matching functions

peb-adr:
@mattbruv if you have no more comments, you can merge now

mattbruv (approved):

mattbruv (approved):

PR #125, opened 2021-01-31
Decomp several zCamera.cpp functions

SquareMan:
Still a noob here so I might be completely off-base but I think I achieved something here.


mattbruv:
Don't forget to remove the decompiled functions from their respective assembly files, otherwise your progress won't get counted on the website!

mattbruv (changes requested):
Remove the decompiled functions from the assembly files

mattbruv (approved):

PR #127, opened 2021-02-01
zSaveLoad.cpp progress

SquareMan:
Several functions here are close but not quite right. Correctly dealing with these larger small functions is still a bit beyond my skillset.

### DIFF
--- a/src/Game/zSaveLoad.cpp
+++ b/src/Game/zSaveLoad.cpp
@@ -2,41 +2,180 @@
 
 #include <types.h>
 
+#include "zGlobals.h"
+#include "zGameState.h"
+#include "../Core/x/xSnd.h"
+#include "../Core/x/xTRC.h"
+#include "../Core/x/xsavegame.h"
+#include "../Core/x/xString.h"
+#include "../Core/x/xEvent.h"
+#include "../CodeWarrior/cstring"
+#include "../CodeWarrior/intrin.h"
+
+extern int8 buffer[16];
+extern int8 buffer[16];
+extern basic_rect<float32> screen_bounds;
+extern basic_rect<float32> default_adjust;
+extern uint32 saveSuccess;
+extern float32 time_elapsed;
+extern float32 time_last;
+extern float32 time_current;
+extern long32 t0;
+extern long32 t1;
+extern int32 currentCard;
+extern int32 currentGame;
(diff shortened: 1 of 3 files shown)


src/Game/zSaveLoad.cpp
```diff
@@ -2,41 +2,180 @@
...
+        }
+    }
+}
+#endif
 
 // func_800AD874
-#pragma GLOBAL_ASM("asm/Game/zSaveLoad.s", "zSaveLoadSGDone__FP17st_XSAVEGAME_DATA")
+void zSaveLoadSGDone(st_XSAVEGAME_DATA* data)
+{
+    if (data->mode == XSG_MODE_LOAD) {
```
mattbruv: Make sure that every C++/Header file that you modify is formatted according to the clang-format rules in the root of this repo. VSCode has a clang-format extension so you can have it run every time you save a file, or you can format them before you commit, but we're trying to keep all of the code formatted consistently

src/Game/zSaveLoad.cpp
```diff
@@ -2,41 +2,180 @@
...
+extern zGlobals globals;
+extern xSndGlobals gSnd;
+extern eGameMode gGameMode;
+extern _tagTRCPadInfo gTrcPad[4];
+extern int32 gGameState;
+extern int32(*xSGT_LoadPrefsCB)(void*, st_XSAVEGAME_DATA*, st_XSAVEGAME_READCONTEXT*, uint32, int32);
+extern int32(*xSGT_LoadLoadCB)(void*, st_XSAVEGAME_DATA*, st_XSAVEGAME_READCONTEXT*, uint32, int32);
+extern int32(*xSGT_SaveProcPrefsCB)(void*, st_XSAVEGAME_DATA*, st_XSAVEGAME_WRITECONTEXT*);
+extern int32(*xSGT_SaveProcCB)(void*, st_XSAVEGAME_DATA*, st_XSAVEGAME_WRITECONTEXT*);
+extern int32(*xSGT_SaveInfoCB)(void*, st_XSAVEGAME_DATA*, int32*, int32*);
```
mattbruv: Put these function declarations in their appropriate header files

mattbruv (changes requested):

mattbruv (approved):

PR #131, opened 2021-02-02
Progress on xCollideFast(1 func), xCutscene(2 funcs), xDecal(1 func and some header stuff) and xEntBoulder(4 funcs)

alexhdevries:


mattbruv (changes requested):
Please remove commented out ASM pragmas as they are not used and could interfere with scraping files

src/Core/x/xCutscene.h
```diff
@@ -92,11 +92,13 @@ struct xCutscene
 };
 
 extern uint32 gFrameCount;
+extern xCutscene sActiveCutscene;
```
mattbruv: All variables should be defined/declared in their C++ files, not header files

alexhdevries: So functions/structs in header files, variables in c++ files?

mattbruv: Yes. Function declarations in header files, function implementations in C++ files, variable declarations in C++ files is the pattern we have stuck to  so far

mattbruv (approved):

PR #134, opened 2021-02-03
more zNPCCommon work

1SDANi:


src/Game/zNPCTypeCommon.cpp
```diff
@@ -504,17 +530,25 @@ void xNPCBasic::PostInit()
 // func_800F4864
 #pragma GLOBAL_ASM("asm/Game/zNPCTypeCommon.s", "__as__10xBaseAssetFRC10xBaseAsset")
 
-// func_800F4888
-#pragma GLOBAL_ASM("asm/Game/zNPCTypeCommon.s", "ImmTranOn__7xPsycheFv")
+void xPsyche::ImmTranOn()
+{
+    flg_psyche = flg_psyche | 1;
```
mattbruv: `flg_psyche |= 1;`

src/Game/zNPCTypeCommon.cpp
```diff
@@ -504,17 +530,25 @@ void xNPCBasic::PostInit()
...
+bool xPsyche::ImmTranIsOn()
+{
+    return flg_psyche & 1;
+}
 
-// func_800F48B4
-#pragma GLOBAL_ASM("asm/Game/zNPCTypeCommon.s", "HasGoal__7xPsycheFi")
+int32 xPsyche::HasGoal(int32 goal)
+{
+    return FindGoal(goal) != NULL;
```
mattbruv, line 550: `return FindGoal(goal);`

mattbruv (changes requested):
Build does not 🆗

src/Game/zNPCTypeCommon.cpp
1SDANi: done

src/Game/zNPCTypeCommon.cpp
1SDANi: that is an invalid line of code, and return (int32)FindGoal(goal); does not ok

mattbruv (changes requested):
Build still doesn't 🆗

PR #142, opened 2021-02-03
even more low hanging zNPCCommon fruit

1SDANi:


mattbruv:
not :OK:

1SDANi:
HOW??? I DOUBLE CHECKED! AGAIN!

src/Game/zNPCTypeCommon.cpp
```diff
@@ -28,22 +54,45 @@ void ZNPC_Common_Shutdown()
...
+    g_skipDescent = 5;
+}
 
-// func_800EEFC0
-#pragma GLOBAL_ASM("asm/Game/zNPCTypeCommon.s", "zNPCCommon_SceneFinish__Fv")
+void zNPCCommon_SceneFinish()
+{
+    zNPCCommon::ConfigSceneDone();
+    NPCSupport_SceneFinish();
+    xDebugRemoveTweak(stringBase0 + 0x42b);
```
mattbruv: Don't name it stringBase0, name it [name of file] + "_strings"

`zNPCTypeCommon_strings`

mattbruv (changes requested):

src/Game/zNPCTypeCommon.cpp
mattbruv: The reason is because every single instance of strings in files is called `stringBase0`, but it doesn't clash because they  are not global scope. When you manually make it global scope, the names would clash if we named them literally.

mattbruv (approved):

PR #180, opened 2021-02-16
Cruise bubble

peb-adr:


src/Game/zEntCruiseBubble.h
```diff
@@ -1,10 +1,98 @@
...
+        // offset 0x24
+        uint32 handle;
+    };
+
+    struct zShrapnelAsset
+    {
+        int32 fassetCount;
+        uint32 shrapnelID;
+        void (*initCB)(zShrapnelAsset*, xModelInstance*, xVec3*, void (*)(zFrag*, zFragAsset*));
+    };
```
seilc: zShrapnelAsset is already defined in zShrapnel.h, it shouldn't be redefined here.

seilc (changes requested):

src/Game/zEntCruiseBubble.h
peb-adr: oops didnt catch that

seilc (approved):

PR #182, opened 2021-02-20
Update build.yml to use BFBB decomp Docker image

mattbruv:


mattbruv:
Build failed because Docker hasn't generated a bfbb image for the new account yet, it will do it whenever there is a push to master.

seilc (approved):

PR #184, opened 2021-02-22
fix bug with symbol ripper

peb-adr:
the ripper tries to guess the subsys name based on the first letter of the symbol and was probably meant as a performance optimization.
Only if the guess cant be made it will search the project files for the symbol.
However the ripper will fail for every symbol beginning with 'x', 'i', or 'z' which isnt following the naming convention of zEnt... or whatever.

Since it doesnt make a notable difference anyway i changed the ripper to always look for the symbol by file search, making it more compatible.

### DIFF
--- a/tools/symbol_ripper/main.py
+++ b/tools/symbol_ripper/main.py
@@ -71,7 +71,7 @@ def cpp_get_base_type(text, unsigned):
 		return rest, "float64"
 	elif char == 'v':
 		return rest, "void"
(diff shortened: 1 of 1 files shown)


peb-adr:
Another few fixes:
- mangled names containing escaped characters are now correctly parsed
- multiple namespace levels for argument types are now correctly parsed
- the ripper is now compatible with multiline pragmas

mattbruv (approved):

PR #226, opened 2021-04-03
Modify diff.py to use TemporaryFile to capture output from make

SquareMan:
On Linux python's `subprocess.run()` seems to have performance issues when capturing stderr using `subprocess.PIPE`. I've updated the script to use `TemporaryFile`s instead and this achieves performance very similar to that under Windows.

I wanted to use `StringIO` first to avoid I/O overhead but `subprocess.run()` requires that the File Object given to it implements `fileno()`. Due to this there may be a small performance hit under Windows, though likely not very noticeable.

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -206,6 +206,14 @@ parser.add_argument(
     "Recommended in combination with -m.",
 )
 parser.add_argument(
+    "-t",
+    "--temporary",
+    dest="temp",
+    action="store_true",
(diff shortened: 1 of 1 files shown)


SquareMan:
I've pushed another commit that adds a new flag `-t` that makes this behavior optional instead of replacing the existing behavior entirely. This eliminates any concern of reducing performance on other systems.

mattbruv (approved):

PR #228, opened 2021-04-03
Start zLasso decomp

JosiahWI:


mattbruv:
Seems like it doesn't 🆗

JosiahWI:
Aha, I forgot to comment out the struct definitions.

mattbruv:
It's good to merge but there are a lot of things to clean up before merging, I'll write it up later

src/Game/zLasso.cpp
```diff
@@ -1,15 +1,174 @@
...
 #include "zLasso.h"
+#include "xEnt.h"
+#include "xSnd.h"
+#include "xstransvc.h"
+#include "xString.h"
+#include "xVec3.h"
 
 #include <types.h>
 
+// lbl_803CDAE0-_SDA2_BASE_(r2), -0x6ea0(r2)
```
mattbruv, line 11: delete these excessive comments above each float extern, they are pretty useless generally

src/Game/zLasso.cpp
```diff
@@ -1,15 +1,174 @@
...
+extern void initVertMap(xModelInstance **a); // parameter should be zLassoGuide *
+extern void bakeMorphAnim(RpGeometry *a, void *b);
+
+typedef struct _y {
+    STRUCT_A Four[3];
+    int pad[2];
+} STRUCT_B;
+
+STRUCT_B gEntities[0xd00]; // lbl_802E5788, -0x7fd2
+#endif
```
mattbruv: I think you should remove all of this junk that's excluded from the file already. It looks more like a madman's notes than helpful documentation or comments. Also, structs in this project are defined in header files and included, not within C++ files.

mattbruv (changes requested):
I'm not sure if you have successfully decompiled any functions yet, but if you have, you should completely delete them from the original `.s` files so that the progress will be updated.

src/Game/zLasso.cpp
JosiahWI: We don't know what these structs are supposed to be. They are placeholders until we figure out the actual data.

mattbruv: How have you surmised what you have now? There isn't any information about this struct in the DWARF data? `gEntities` isn't a global variable in the `symbols.txt`, which means it doesn't exist in the source code, so I'm not sure what prompted you to make it.

src/Game/zLasso.cpp
```diff
@@ -21,22 +180,159 @@
...
+    lasso->reindex[0] = 0;
+    lasso->reindex[1] = 1;
+    lasso->reindex[2] = 2;
+    lasso->reindex[3] = 3;
+    lasso->reindex[4] = 4;
+}
+#endif
 
 // func_8009D8D4
+#if 0
```
SquareMan, line 221: This function should have the `#if 0` statement and `#pragma GLOBAL_ASM` statement removed if it's matching. Additionally the assembly for it should be removed from the .text section of `zLasso.s` so that the progress tracker will count it correctly

SquareMan (changes requested):
Adding this to what @mattbruv already wrote.

PR #248, opened 2021-06-15
progress?? :eyes: 

peb-adr:
did a few functions on a train ride

### DIFF
--- a/asm/Game/zNPCHazard.s
+++ b/asm/Game/zNPCHazard.s
@@ -755,11 +755,8 @@ lbl_801881F8:
 /* 80188204 00185004  38 21 00 10 */	addi r1, r1, 0x10
 /* 80188208 00185008  4E 80 00 20 */	blr 
 
-/* changed from ... */
-/* HAZ_Iterate__FPFR9NPCHazardPv_bPvi */
-/* ... so linker can find it */
-.global HAZ_Iterate__FPFR9NPCHazardPv_UcPvi
-HAZ_Iterate__FPFR9NPCHazardPv_UcPvi:
+.global HAZ_Iterate__FPFR9NPCHazardPv_bPvi
+HAZ_Iterate__FPFR9NPCHazardPv_bPvi:
 /* 80188218 00185018  94 21 FF E0 */	stwu r1, -0x20(r1)
 /* 8018821C 0018501C  7C 08 02 A6 */	mflr r0
 /* 80188220 00185020  3C C0 80 36 */	lis r6, g_hazards@ha
--- a/asm/Game/zNPCHazard.s
+++ b/asm/Game/zNPCHazard.s
@@ -755,8 +755,11 @@ lbl_801881F8:
 /* 80188204 00185004  38 21 00 10 */	addi r1, r1, 0x10
 /* 80188208 00185008  4E 80 00 20 */	blr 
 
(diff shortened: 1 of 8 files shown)


asm/Game/zNPCHazard.s
```diff
@@ -755,8 +755,11 @@ lbl_801881F8:
 /* 80188204 00185004  38 21 00 10 */	addi r1, r1, 0x10
 /* 80188208 00185008  4E 80 00 20 */	blr 
 
-.global HAZ_Iterate__FPFR9NPCHazardPv_bPvi
-HAZ_Iterate__FPFR9NPCHazardPv_bPvi:
+/* changed from ... */
+/* HAZ_Iterate__FPFR9NPCHazardPv_bPvi */
+/* ... so linker can find it */
+.global HAZ_Iterate__FPFR9NPCHazardPv_UcPvi
+HAZ_Iterate__FPFR9NPCHazardPv_UcPvi:
```
seilc: The linker couldn't find HAZ_Iterate because its declaration in zNPCHazard.h was incorrect (DWARF treats bool as uint8 for whatever reason).

`void HAZ_Iterate(uint8 (*fp)(NPCHazard&, void*), void* context, int32 flag_filter);` should be changed to:
`void HAZ_Iterate(bool (*fp)(NPCHazard&, void*), void* context, int32 flag_filter);`

Then change the HAZ_Iterate symbol back here and in zNPCHazard.cpp, and the linker should be able to find it.

src/Game/zEntCruiseBubble.cpp
```diff
@@ -452,9 +459,9 @@ void cruise_bubble::start_damaging()
 
 // func_80057684
 #ifndef NONMATCHING
-#pragma GLOBAL_ASM("asm/Game/zEntCruiseBubble.s", "damage_entity__Q213cruise_bubble30_esc__2_unnamed_esc__2_zEntCruiseBubble_cpp_esc__2_FR4xEntRC5xVec3RC5xVec3RC5xVec3fb")
+#pragma GLOBAL_ASM("asm/Game/zEntCruiseBubble.s", "damage_entity__13cruise_bubbleFR4xEntRC5xVec3RC5xVec3RC5xVec3fb")
 #else
-void cruise_bubble::damage_entity(xEnt& ent, const xVec3& loc, const xVec3& dir, const xVec3& hit_norm, float32 radius, uint8 explosive)
+void cruise_bubble::damage_entity(xEnt& ent, const xVec3& loc, const xVec3& dir, const xVec3& hit_norm, float32 radius, bool explosive)
```
seilc: I think this is supposed to be an unnamed namespace within the cruise_bubble namespace.

```c++
namespace cruise_bubble
{
    namespace
    {
        void damage_entity(xEnt& ent, const xVec3& loc, const xVec3& dir, const xVec3& hit_norm, float32 radius, bool explosive)
        // ...
```

Try that and see if the symbol matches.

seilc (changes requested):

peb-adr:
Ok so this proofs the namespace thing can be fixed.
Doing it for the rest of the file has some annoying implications so, ill go about that when im back at my pc, rather then on laptop.
Also the build checker seems to have broken over night. It clean builds OK for me tho.

seilc (approved):

PR #255, opened 2022-01-06
Add script for building with Docker

SquareMan:
This script should provide an easy option for building on linux without having to worry about getting a working Wine version setup and devkitPro installed. I've even added a new flag to the diff script to using this build script instead of make directly. The diff script is much less responsive this way, but I think it's because the container we're using does not have the wineprefix preconfigured, so it has to do it each time.


SquareMan:
This will require an updated docker image to be pushed for the build script to work as-is, so hold off on merging for now. If we decide that we actually don't want a build script to execute make in docker then we probably want to add the necessary tools and python libs to the container to at least be able to run things like `diff.py` in the container.

PR #256, opened 2022-09-17
Update Dockerfile to pin DKP

mkst:


SquareMan (approved):
LGTM, thanks for fixing the github actions hack as well.

PR #257, opened 2022-12-10
Fully redisassemble repo

encounter:
- Now works with devkitPro r40+
- Fully split with all visibility info, symbols and types from original .elf
- Fully shiftable
- Various improvements to build system

I integrated my project [decomp-toolkit](https://github.com/encounter/decomp-toolkit), which is now used to fix `powerpc-eabi-as` object files to work correctly with `mwldeppc`. It also provides `sha1sum` and `elf2dol`, so those now use `dtk`.

Things left to do:
- [ ] Re-organize new asm files into directories
- [ ] Split `COM` symbols, they're currently stuffed at the end of one of the files.
  - They'll be moved to individual files in the form of `.comm name, size, align` (align will be 4)
- [ ] Update compiler flags
  - Use `-common on` for main source files (global BSS variables are pushed to the end of `.bss` and deduplicated)
  - Use `-sym on` for main source files (affects function ordering & generates multiple `.text` sections)
- [ ] Verify matches for src files and re-link them (objdiff will help)
- [ ] Fix progress calculation


SquareMan:
Really good stuff here, thanks for the PR. I think this will help a lot with getting this project moving again. I'm curious about what some of the new ASM files are for (renderware?) but not going to look too closely until the PR is ready for review.

encounter:
Not sure yet, all the unit names were extracted from the .elf, which doesn’t provide anything more than just the filename.

I could use help from project maintainers to finish the to-do list. In particular, assisting with organizing the asm and updating src files to match.

mattbruv:
Yes, I will work on your TODO list and try to help. This week is going to be very chaotic for me as I am wrapped up in traveling but once I'm back in the US I should have a good amount of time to sit down and get things organized over the holidays

mattbruv:
I get an error during the make process:

````
\bfbb> make
 AS      asm/xAnim.s
asm/xAnim.s: Assembler messages:
asm/xAnim.s:4477: Error: unknown pseudo-op: `.bss'
make: *** [Makefile:152: obj/asm/xAnim.o] Error 1
```

encounter:
It's likely you'll need to update to the latest devkitPro.

mattbruv:
I got it to build `OK` by running this script first:

```py
for a in asms:
    f = open(a).read()
    text = f.replace(".bss\n", ".section .bss\n")
    open(a, "w").write(text)
```

Is there a special reason why that particular section wasn't prefixed with `.section`? If we can add it without it hurting anything in the later versions, I think we should because it would support older versions of DevkitPro as well

encounter:
I was just using the built-in directives where possible, it generates slightly better output from the assembler. But my elf fixup step already unifies the output, so it’s fine to use either.

mattbruv:
Nice, your most recent fixes work fine for me and it also now passes the CI build.

I'm wondering what our goal should be before we merge this. This undoubtedly needs to become our working branch now. We can merge it now and work on integrating our C++ code back into it, or we can figure it all out before merging it to master. @seilweiss thoughts?

I don't think it particularly matters which order we do it in. One thing I want to do is have a similar workflow that will enable us to decompile functions out of order. Now instead of inlining the raw bytes like the old way we could have a similar pre-processing script but this time it actually patches in the real ASM instructions in the C++ source. I really like keeping as much junk out of our CPP source files as possible

SquareMan:
Do we need to remove the decomped functions from the ASM again or is progress tracked differently now?

encounter:
> Do we need to remove the decomped functions from the ASM again or is progress tracked differently now?

I forgot to mention this in the TODOs as well, this breaks the existing method of calculating progress. I'm attempting to figure a better solution.

mattbruv:
Don't worry about progress tracking for now. There's probably a better method to deal with that, and I will handle that once we get this branch ready and merged.

What's more important is to get this thing ready and to get us moving again in decompiling code.

> Re-organize new asm files into directories

This is only for organizational/aesthetic purposes right? If so, I don't think it's really necessary. Personally I'm fine with just having all of the ASM files dumped into a folder. They are already separated into their correct files. I think we can skip this part (unless someone wants to take the time to do this, then sure).

**EDIT: it looks like objdiff needs the files to be in matching directories to load correctly**

> Split COM symbols

Can you please give more context to this one? I'm trying to find an example of COM in the ASM files but it's impossible to find because COM is used so much in the codebase

https://stackoverflow.com/a/501138/2936448

JustOlaia:
""LINK    main.elf
Identifier "x\0A\0n\0i\0m\0.\0o\0" was not in the list of identifiers passed to GnuBuilder::new()
make: *** [Makefile:145: main.elf] Error 1""
i got this error

encounter:
That’s fixed in dtk v0.2.3, update dtk_version file

JustOlaia:
> That’s fixed in dtk v0.2.3, update dtk_version file

do i download "dtk-windows-x86_64.exe" or "dtk-windows-arm64.exe"?

JustOlaia:
thanks it works now

JustOlaia:
well it seems like it didnt do nothing with the src/game code or im just dumb

mattbruv:
This branch isn't configured to build our C++ source yet. It is a 100% assembly build. We are still in the process of getting this branch to work with our C++ source

JustOlaia:
oh ok

mattbruv:
I'm getting the following error when building this after a fresh install, not sure what exactly is going on:

dtk.exe 0.2.2 828766b22b170b8863b72ddd1dcaab99455801b2

>  LINK    main.elf
Identifier "x\0A\0n\0i\0m\0.\0o\0" was not in the list of identifiers passed to GnuBuilder::new()
make: *** [Makefile:145: main.elf] Error 1

encounter:
> That’s fixed in dtk v0.2.3, update dtk_version file

mattbruv:
My mistake, I should pay more attention to the comments in this thread 😅 Sorry.

mattbruv (approved):
@encounter We are forever in your debt 🙏

mattbruv:
Merging this now to work off of this as our main branch and will fix the issues as we run into them.

PR #278, opened 2024-06-25
Finish decompiling zTaxi & mark it as equivalent

tgsm:


mattbruv (approved):

src/SB/Game/zTalkBox.h
```diff
@@ -48,7 +48,8 @@ struct ztalkbox : xBase
     {
         ANSWER_CONTINUE,
         ANSWER_YES,
-        ANSWER_NO
+        ANSWER_NO,
+        ANSWER_3,
```
SquareMan, line 52: Should probably just call this ANSWER_NUM, though it's somewhat confusing to me that this enum definition in the dwarf doesn't have this value. Possibly sAnswer was just defined as `ztalkbox::ANSWER_NO + 1` but that's also pretty strange.

PR #281, opened 2024-06-27
Finish matching xSurface

tgsm:


src/SB/Core/x/xSurface.cpp
```diff
@@ -25,32 +23,22 @@ void xSurfaceInit(uint16 num_surfs)
     }
 }
 
-#endif
+extern "C" {
+extern void __copy(void* dst, const void* src, uint32 n);
+}
```
SquareMan, line 29: Should this be declared in intrin.h?

tgsm: I'm not sure. It doesn't get replaced with an instruction like the other intrinsics do, and it's implemented in the runtime library.

SquareMan: I don't know either. I don't think it's something that's supposed to be directly included though. The whole assignment function was probably compiler generated, with the default implementation being to call `__copy` for any arrays. I suppose this fine for now.

PR #299, opened 2024-07-08
Start decompiling xHud

tgsm:


src/SB/Core/x/xHud.h
```diff
@@ -11,6 +11,25 @@ typedef struct widget;
 
 namespace xhud
 {
+    struct block_allocator
+    {
+        U32 unk0;
+        U32 unk4;
+        void* unk8;
+        block_allocator* next;
```
SquareMan: it looks like these fields are defined in the dwarf data:
```cpp
    unsigned int _block_size; // offset 0x0, size 0x4
    unsigned int _alloc_size; // offset 0x4, size 0x4
    class holder * _top; // offset 0x8, size 0x4
    class block_allocator * _next_alloc; // offset 0xC, size 0x4
```

PR #301, opened 2024-07-09
zNPCFXCinematic

TotallyNotMichael-GH:
My first actual bit of work. VERY excited!


mattbruv (commented):
Need to format these files according to our `.clang-format` file rules. If you enable autoformatting on save in your VS code settings it should do this for you automatically.

mattbruv (approved):

PR #304, opened 2024-07-10
Got some more size 4 voids out of the way

TotallyNotMichael-GH:


TotallyNotMichael-GH:
having pull request issues, not really sure what to do.

src/SB/Core/x/xShadow.h
```diff
@@ -33,3 +33,11 @@ void xShadowManager_Reset();
 void xShadowManager_Render();
 
 #endif
+
+void xDrawSphere(const xVec3*, F32, U32)
```
SquareMan: this function definition is causing your ci errors (run `ninja all_source` to test locally). I think it's supposed to be a WEAK function. check if the symbol has a `w` icon before it in objdiff.

SquareMan (changes requested):

src/SB/Core/x/xShadow.h
SquareMan: There might be other WEAK functions in this TU too so look for that.

src/SB/Core/gc/isavegame.cpp
```diff
@@ -107,3 +107,11 @@ void iSGSessionEnd(st_ISGSESSION* isgdata)
     memset(isgdata, 0, sizeof(st_ISGSESSION));
 }
 #endif
+
+void iSG_cb_asyndone(long, long)
```
mattbruv, line 111: These should be S64

src/SB/Core/x/xScrFx.cpp
```diff
@@ -1,3 +1,11 @@
 #include "xScrFx.h"
 
 #include <types.h>
+
+void xScrFxDistortionAdd(xVec3*, xVec3*, int)
```
mattbruv: S32

src/SB/Core/x/xTRC.cpp
```diff
@@ -1,3 +1,7 @@
 #include "xTRC.h"
 
 #include <types.h>
+
+void xTRCPad(int, _tagTRCState)
```
mattbruv: S32

mattbruv (changes requested):

src/SB/Core/gc/isavegame.cpp
SquareMan: I'm not sure about that, they're definitely just `long`s in the mangled name. I think this one function needs to be reverted back to how it was because we don't have a typedef for `long` and it mangles differently from `int`.

If you review the discord discussion from before seil updated the types it looks like the platform specific code uses `long`s in some places so it may have been another typedef they had that was differently sized depending on platform but I think we're just using `long` directly for now.

PR #307, opened 2024-07-11
First batch of Size 8s

TotallyNotMichael-GH:


src/SB/Game/zCameraTweak.h
```diff
@@ -3,6 +3,10 @@
 
 #include "xDynAsset.h"
 
+static float sCamPitch;
+static float sCamH;
+static float sCamD;
```
mattbruv: move these to zCameraTweak.cpp

mattbruv (changes requested):

mattbruv (approved):

PR #344, opened 2024-08-04
Decompile most of zNPCGlyph

tgsm:


src/SB/Game/zNPCGlyph.cpp
```diff
@@ -23,14 +122,492 @@ void zNPCGlyph_ScenePostInit()
...
+}
+
+void NPCGlyph::VelSet_Shiny()
+{
+    tmr_glyph = -1.0f;
+    xVec3 local_18 = {};
+
+    {
+        xVec3 temp;
+        // Do we use sph here?
```
SquareMan, line 509: I've been using cylinder for the player but I'm not sure if that's actually correct yet.

PR #355, opened 2024-08-20
Various matches

escape209:
`ZNPC_AnimTable_BossSBobbyArm` improved to _nearly_ perfect.


escape209:
Don't merge yet, I just noticed an issue 🤦

escape209:
Fixed the issue, though removing the non-matching ifdef had an effect on stringbase that I didn't notice, so readding non-matching til that's resolved (also, it's technically still non-matching, anyway.)

PR #366, opened 2024-09-11
zLightning work

escape209:
Size of `sLightning` object contradicts DWARF for whatever reason (version difference?).


mattbruv:
CI failed

PR #378, opened 2024-09-29
zPlatform matches + fix typo

escape209:


escape209:
ffs
1 sec

escape209:
Ok we're good

escape209:
Gonna check if I can get rid of the intrinsics: if not, then good to go

escape209:
Okay actually good to go, finally

PR #391, opened 2024-10-11
Match zMusicUnpause

escape209:
Thanks @tgsm
Also remove outdated comment and match getCurrLevelMusicEnum linkage
Tried to find more stuff to put in here but it's a tough file 💀

### DIFF
--- a/src/SB/Game/zMusic.cpp
+++ b/src/SB/Game/zMusic.cpp
@@ -114,8 +114,7 @@ void zMusicInit()
     volume_reset();
 }
 
-// Correct, but won't work due to the switch case jump table (messes with offsets)
-S32 getCurrLevelMusicEnum()
+static S32 getCurrLevelMusicEnum()
 {
     S32 snd_enum;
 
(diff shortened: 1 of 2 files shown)


src/SB/Game/zMusic.h
```diff
@@ -38,7 +38,7 @@ struct zMusicTrackInfo
 
 void volume_reset();
 void zMusicRefreshVolume();
-S32 getCurrLevelMusicEnum();
+static S32 getCurrLevelMusicEnum();
```
SquareMan: This function declaration should be removed from the header since other cpp files won't be able to call this function

src/SB/Game/zMusic.cpp
```diff
@@ -114,7 +114,6 @@ void zMusicInit()
     volume_reset();
 }
 
-// Correct, but won't work due to the switch case jump table (messes with offsets)
 S32 getCurrLevelMusicEnum()
```
SquareMan: Looks like this should be static since you changed the prototype?

SquareMan (changes requested):

escape209:
Integrated both of the changes

PR #396, opened 2025-02-03
zEntDestructObj: SwapModel 100% match

escape209:


src/SB/Game/zEntDestructObj.cpp
```diff
@@ -28,9 +28,9 @@ namespace
             RwMatrixCopyMacro(modelInst->Mat, ent->model->Mat);
             
             modelInst->Flags &= 0xBBFC;
-            modelInst->Flags |= (s16) (ent->model->Flags & 0x4403);
+            modelInst->Flags |= (u16) (ent->model->Flags & 0x4403);
             ent->model = modelInst;
-            iBoxForModelLocal((xBox*) &ent->bound.box, (xModelInstance *) modelInst);
+            iBoxForModelLocal((xBox*) &ent->bound.box.box, (xModelInstance *) modelInst);
```
SquareMan: We can get rid of the cast to `xBox*` now I think.

SquareMan (changes requested):

src/SB/Game/zEntDestructObj.cpp
SquareMan: While you're at it, it looks like `modelInst` doesn't need to be casted either.

escape209: Done

PR #401, opened 2025-02-10
zEntSimpleObj - Most Functions to 100% Code Match

JoshSanch:
Mgr functions are tough - will revisit those at a later time.


src/SB/Game/zEntDestructObj.cpp
```diff
@@ -48,141 +53,176 @@ void zEntDestructObj_FindFX()
...
     u32 sfxHitId;
     u32 hitModelId;
     u32 destroyModelId;
     U8 boundType;
-    xModelInstance *destroyModel;
-    xModelInstance *hitModel;
+    xModelInstance* destroyModel;
+    xModelInstance* hitModel;
+
+    zEntInit((zEnt*)ent, asset, 0x44535452U);
```
SquareMan, line 79: this int literal should be a multicharacter literal.
`zEntInit((zEnt*)ent, asset, 'DSTR');` I think.

src/SB/Game/zEntSimpleObj.cpp
```diff
@@ -2,21 +2,727 @@
...
+        {
+            RwFree(temp_r27);
+            return;
+        }
+        sMgrCount = var_r26;
+        sMgrList = (xSphere*)xMemAlloc(gActiveHeap, var_r26 << 6, 0x40);
+        var_r29_2 = temp_r27;
+        var_r30_2 = 0U;
+        var_r25_2 = sMgrList;
+        // while (var_r30_2 < var_r26)
```
SquareMan, line 189: What's all this code? Is this function WIP? No change needed if that's the case but I just want to make sure this isn't something you forgot to delete.

SquareMan (changes requested):

PR #407, opened 2025-02-16
zNPCSpawner WIP

daft7:
Added several methods with partial code match to zNPCSpawner and some light struct edits to incorporate these.

### DIFF
--- a/src/SB/Game/zNPCSpawner.cpp
+++ b/src/SB/Game/zNPCSpawner.cpp
@@ -177,3 +177,94 @@ U8 zMovePoint::IsOn()
 {
     return this->on;
 }
+
+
+st_XORDEREDARRAY* zNPCSpawner::FillPending()
+{
+    ClearPending();
+    ReFillPending();
+    return this->actvlist; 
+}
+
+st_XORDEREDARRAY* zNPCSpawner::ReFillPending()
+{
+    s32 var_r28;
+    zNPCCommon *temp_r29;
+    zNPCSpawner *var_r30;
+
+    var_r28 = 0;
+    var_r30 = this;
+    do {
+        temp_r29 = var_r30->npc_owner;
+        if (((zNPCCommon*)var_r30->npc_owner != NULL) && ((s32) temp_r29->pad0[4] == 1))
+        {
+            XOrdAppend(this->pendlist, (void*) temp_r29);
+            temp_r29->pad0[4] = 2;
+        }
+        var_r28 += 1;
+        var_r30 += 0xC;
+    } while (var_r28 < 0x10);
+    return this->actvlist;
+}
+
+
+// void zNPCSpawner::ClearActive()
+// {
+//     s32 var_r6;
+//     s32 var_r7;
+//     void *temp_r5;
+
+//     var_r7 = 0;
+//     var_r6 = 0;
+// loop_4:
+//     if (var_r7 < (s32) this->cnt_cleanup) {
+//         temp_r5 = *(this->pendlist->list + var_r6);
+//         if (temp_r5 != NULL) {
+//             (u32) temp_r5[1] = 1;
+//         }
+//         var_r6 += 4;
+//         var_r7 += 1;
+//         goto loop_4;
+//     }
+//     XOrdReset__FP16st_XORDEREDARRAY(&this->unk1B0);
+// }
+
+SMNPCStatus* zNPCSpawner::NextPendingNPC( s32 arg0 )
+{
+    S32 temp_r4;
+    const F32* temp_ptr = NULL;
+
+    temp_r4 = (S32) this->actvlist;
+    if (temp_r4 < 1) {
+        return NULL;
+    }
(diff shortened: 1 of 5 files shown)


src/SB/Game/zNPCSpawner.h
```diff
@@ -50,7 +51,10 @@ enum en_SM_NPC_STATUS
     SM_NPC_FORCE = 0x7fffffff
 };
 
-struct zNPCCommon;
+struct zNPCCommon {
+    /* 0x000 */ char pad0[0x1B8];
+    /* 0x1B8 */ void *unk1B8;                       /* inferred */
+};                                                  /* size >= 0x1BC */
```
SquareMan: This struct is defined in zNPCTypeCommon.h, you should be able to use that.

src/SB/Game/zNPCSpawner.h
```diff
@@ -83,8 +87,8 @@ struct zNPCSpawner : RyzMemData
     S32 cnt_spawn;
     SMSPStatus sppool[16];
     SMNPCStatus npcpool[16];
-    st_XORDEREDARRAY pendlist;
-    st_XORDEREDARRAY actvlist;
+    st_XORDEREDARRAY* pendlist;
+    st_XORDEREDARRAY* actvlist; //0x1A4
```
SquareMan: Can you explain why this change was necessary. The struct definition came from the ps2 dwarf and this struct would be unlikely to make a change like this across platforms.

src/SB/Game/zNPCSpawner.cpp
```diff
@@ -70,8 +70,8 @@ void zNPCSpawner::Subscribe(zNPCCommon* owner)
     this->tym_delay = _805_Spawner;
     this->max_spawn = -1;
     this->wavestat = SM_STAT_BEGIN;
-    XOrdInit(&this->pendlist, 0x10, 0);
-    XOrdInit(&this->actvlist, 0x10, 0);
+    XOrdInit(this->pendlist, 0x10, 0);
+    XOrdInit(this->actvlist, 0x10, 0);
 }
```
SquareMan: It looks like this function matched before. I understand that you changed this because you changed the struct def but surely this doesn't match anymore since you're passing the value rather than the address of the value right?

SquareMan (changes requested):

src/SB/Game/zNPCSpawner.h
SquareMan: Looks like this having this redefinition is breaking the build too.

daft7: Updated to use definition from zNPCTypeCommon along with some fixes to avoid recursive imports.

src/SB/Game/zNPCSpawner.h
daft7: Reverted in latest commit.

src/SB/Game/zNPCSpawner.cpp
daft7: I changed it back now.

src/SB/Game/zNPCSpawner.cpp
```diff
@@ -177,3 +177,95 @@ U8 zMovePoint::IsOn()
...
+    {
+        temp_r29 = var_r30->npc_owner;
+        if (((zNPCCommon*)var_r30->npc_owner != NULL) && ((s32)temp_r29->flg_vuln == 1))
+        {
+            XOrdAppend(&this->pendlist, (void*)temp_r29);
+            temp_r29->flg_vuln = 2;
+        }
+        var_r28 += 1;
+        var_r30 += 0xC;
+    } while (var_r28 < 0x10);
```
SquareMan, line 206: This is probably some sort of for loop over an array, with `var_r29` and `var_r30` being compiler generated variables related to indexing an array.

src/SB/Game/zNPCSpawner.cpp
```diff
@@ -177,3 +177,95 @@ U8 zMovePoint::IsOn()
...
+    return &this->actvlist;
```
SquareMan, line 207: This is probably some sort of for loop over an array, with `var_r29` and `var_r30` being compiler generated variables related to indexing an array.

src/SB/Game/zNPCSpawner.cpp
```diff
@@ -177,3 +177,95 @@ U8 zMovePoint::IsOn()
...
+    s32 var_ctr;
+    SMNPCStatus* var_r6;
+    zNPCCommon* temp_r0;
+    zNPCCommon* temp_r0_2;
+    zNPCCommon* temp_r0_3;
+    zNPCCommon* temp_r0_4;
+    zNPCCommon* temp_r0_5;
+    zNPCCommon* temp_r0_6;
+    zNPCCommon* temp_r0_7;
+    zNPCCommon* temp_r0_8;
```
SquareMan, line 256: eventually all these unused variables from decompiler output should be cleaned up

SquareMan (approved):
LGTM. Welcome Aboard!

Left a few nits on some stuff since some of it is still very "decompiler-esque" but not gonna block merging since this code is all very WIP still anyways.

PR #409, opened 2025-02-17
Add full headers for dolsdk2004

SquareMan:
Basically just the full set of headers from [doldecomp/dolsdk2004](https://github.com/doldecomp/dolsdk2004)

This may not be 100% accurate for us but I figure that we will need to edit these headers much less often than we currently need to edit our monolithic dolphin.h

Notable Changes:

# fastmath.h

Starting getting a redefinition error for `Vec` in xVec3.cpp for some reason after this. It appears to be unused so I just deleted it from the fastmath header.

# dolphin/types.h

adjusted std includes to use our std headers.

# dolphin/card.h
renamed two function arguments from `new` to `_new` to avoid conflicting with the c++ keyword

Added missing defines:
```
#define CARD_STAT_ANIM_LOOP 0x00
#define CARD_STAT_ANIM_BOUNCE 0x04
#define CARD_STAT_ANIM_MASK 0x04
```


JoshSanch:
We are merging this PR in, acknowledging that these headers will be an inexact match for this project. However, they will be closer to what we need than what we have now, so the progress is welcome. Thanks for the effort @SquareMan

PR #410, opened 2025-02-19
xMath3_WIP

daft7:
Groundwork for xMath3.

### DIFF
--- a/src/SB/Core/x/xClimate.h
+++ b/src/SB/Core/x/xClimate.h
@@ -38,4 +38,7 @@ void xClimateSetSnow(F32 stre);
 void xClimateInitAsset(_tagClimate* climate, xEnvAsset* easset);
 void xClimateInit(_tagClimate* climate);
 
+void xMat3x3Identity(xMat3x3* matrix);
+
+
 #endif
--- a/src/SB/Core/x/xClimate.h
+++ b/src/SB/Core/x/xClimate.h
@@ -39,6 +39,6 @@ void xClimateInitAsset(_tagClimate* climate, xEnvAsset* easset);
 void xClimateInit(_tagClimate* climate);
 
 void xMat3x3Identity(xMat3x3* matrix);
-void xMat3x3Copy(xMat3x3* m1, const xMat3x3* m2);
+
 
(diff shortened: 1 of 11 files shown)


src/SB/Core/x/xClimate.h
```diff
@@ -38,4 +38,7 @@ void xClimateSetSnow(F32 stre);
 void xClimateInitAsset(_tagClimate* climate, xEnvAsset* easset);
 void xClimateInit(_tagClimate* climate);
 
+void xMat3x3Identity(xMat3x3* matrix);
+void xMat3x3Copy(xMat3x3* m1, const xMat3x3* m2);
+
```
SquareMan: I think this is your build error, because `xMat3x3` isn't declared yet here. Regardless these functions are declared in xMath3.h already so you can remove these and just include xMath3.h whenever you need to use them (even though the functions themselves are in xClimate.cpp)

src/SB/Core/x/xMath3.h
```diff
@@ -125,7 +125,7 @@ void xMat3x3Mul(xMat3x3* o, const xMat3x3* a, const xMat3x3* b);
 void xMat3x3SMul(xMat3x3*, const xMat3x3*, F32);
 void xBoxFromLine(xBox& box, const xLine3& line);
 void xBoxFromRay(xBox& box, const xRay3& ray);
-void xMat3x3Identity(xMat3x3* matrix);
+void xMat3x3Identity(xMat3x3* matrix); // May come from xClimate
```
SquareMan: A bit of a nit I know, but can you remove these comments (or at least reword them somehow). They don't really "come from" xClimate, but rather would have been defined here directly in this header file and just by chance the definition in xClimate's translation unit ended up being the one that the linker chose to use (maybe it was the first/last usage or something like that).

SquareMan (changes requested):

src/SB/Core/x/xMath3.h
SquareMan: If you want to leave a comment about this I would probably say something like
`// TODO: These functions should be inline`

PR #413, opened 2025-02-21
Un-extern some strings in zLightningTweakChangeType

joeyballentine:
Honestly, not sure if this is correct, but it got the function's matching score up a little bit. Anyone have any idea why those `+ 0x18`s are necessary?

### DIFF
--- a/src/SB/Game/zLightning.cpp
+++ b/src/SB/Game/zLightning.cpp
@@ -12,16 +12,13 @@ extern RwRaster* sLightningRaster;
 extern xVec3 sTweakStart;
 extern xVec3 sTweakEnd;
 
-extern char stringBase0[];
-
 void xDebugAddTweak(const char*, F32*, F32, F32, const tweak_callback*, void*, U32)
 {
 }
 
 void lightningTweakChangeType(const tweak_info& t)
 {
-    // "Lightning|\x01Type Info"
-    xDebugRemoveTweak(stringBase0 + 0x18);
+    xDebugRemoveTweak("Lightning|\x01Type Info" + 0x18);
 
     switch (gLightningTweakAddInfo.type)
     {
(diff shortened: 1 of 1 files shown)


JoshSanch:
Could be that it's trying to start at a specific offset in the string data.

Might be better expressed as:
`"Lightning|\x01Type Info|Setup Degrees"[0x18]`

joeyballentine:
I'm almost willing to bet that this isn't what they wrote, and that it has something to do with reusing the string `Lightning|\x01Type Info` for all these other strings via string concatenation, since the string is 23 characters long and 0x18 is 24. However, I tried doing that and it made it match even less. I'm open to any other suggestions though

JoshSanch:
Have you tried just using the string at that starts at the offset? Not something I've ever confirmed but I would think the way it ends up in the program memory, it's like using a specific substring that gets loaded in from the static data. So like:
` "Lightning|\x01Type Info|Setup Degrees"[0x18]`
-->
` "Setup Degrees"`

joeyballentine:
The problem with that is that the `addi`s end up incorrect

image

joeyballentine:
This is how it compiles the way i have it 
image

SquareMan:
So strings are treated specially by the compiler. All the strings used in a single translation unit get pooled together into the single `stringBase` symbol. Then to access a given string literal the compiler will reference the string base symbol and add the offset required to get to that specific string in the pool.

This was originally old, pre dtk code. To get matching assembly a fake extern char* symbol was used and the offset was manually added.

Tldr the correct string is likely the literal you have now without any of the weird indexing stuff. This may not match for the stringBase offset if previous functions that use strings aren't decomped but what's important is that the string itself is correct. You can verify this by viewing the stringBase symbol data and seeing what string starts at the offset used.

Additionally recent versions of objdiff should show the specific string used if you hover the instruction that adds the offset.

joeyballentine:
Oh, good to know. I'll verify that and remove the offsets

PR #416, opened 2025-02-24
Revert '!' back to hex (0x21d)

joeyballentine:
Turns out not everything that's hex and *can* be converted to a char, *should*. 0x21 is !, 0x21d is nonsense. Also that argument is called "line" and it's an int, so i was just being dumb thinking this could be converted like the other one that was "HUB2" in hex.

My bad.

I do wonder: How did this `OK` when that was clearly wrong?


mattbruv (approved):

SquareMan:
> How did this OK when that was clearly wrong?

iSystem currently isn't linked the matching build since it isn't finished.

PR #438, opened 2025-03-16
Restore Partial/Nearly Complete Matches Projectwide

JoshSanch:
This PR restores many functions that were macro'd away as a result of lack of support from the project's former tooling to support building with partial matches. This is not a concern in the modern era, so I've restored all of these functions so that the project's progress tracker and objdiff comparisons are up-to-date with real progress.

Also includes a few minor float data substitutions and some other very small fixes.

Confirmed all files build using `ninja all_source`.


mattbruv (approved):
This is a chonker

PR #445, opened 2025-03-18
zNPCTypeDutchman: added Death::Enter function

RejectSanity:


src/SB/Game/zNPCTypeDutchman.cpp
```diff
@@ -207,6 +207,20 @@ S32 zNPCGoalDutchmanCaught::Exit(float dt, void* updCtxt)
...
 
+
+
+
+S32 zNPCGoalDutchmanDeath::Enter(float dt, void* updCtxt)
+{
+    owner.delay = 0.0f;
+    zNPCGoalCommon::Enter(dt, updCtxt);
+
+    return;
```
tgsm:
```suggestion
    return zNPCGoalCommon::Enter(dt, updCtxt);
```

src/SB/Game/zNPCTypeDutchman.cpp
```diff
@@ -207,6 +207,20 @@ S32 zNPCGoalDutchmanCaught::Exit(float dt, void* updCtxt)
     return xGoal::Exit(dt, updCtxt);
 }
 
+
+
+
+S32 zNPCGoalDutchmanDeath::Enter(float dt, void* updCtxt)
```
mattbruv: The `float` should be changed to `F32` to match our types (which were how Heavy Iron defined their types in the original codebase)

mattbruv (changes requested):

src/SB/Game/zNPCTypeDutchman.h
```diff
@@ -309,6 +309,7 @@ struct zNPCGoalDutchmanDeath : zNPCGoalCommon
     F32 emit_frac;
     F32 min_y;
     F32 max_y;
+    S32 Enter(float, void*);
```
mattbruv: Not trying to be pedantic, but we also want to change `float` to `F32` here in the function definitions/headers files too. We also want to try and keep the parameter names in the headers as well if we know them, for consistency's sake. So adding `dt` and `updCtxt` as well in the headers.

PR #454, opened 2025-03-19
Add DWARF debug output based on dtk 1.4.0

escape209:
Non-zero chance my hacky splitting script might have screwed up somewhere, but don't think so (based on a semi-brief glance.)


JoshSanch:
I'm not sure that we should delete the original DWARF file until we have a way of 100% guaranteeing that your script outputs a 100% match here. It'd be better to update our tooling to ignore the file.

joeyballentine (approved):
Josh says it's good so I'm gonna merge this

JoshSanch:
don't even have write perms B)

mattbruv:
Remind me tomorrow and I'll get you write permission

PR #482, opened 2025-03-27
Update Github CI to Treat Warnings as Errors and Check Non-Matching Files for Errors

JoshSanch:
Fixes a couple of warnings that were preventing builds from succeeding as a result of the build flag changes.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -73,6 +73,14 @@ jobs:
           name: ${{ matrix.version }}_previous
           path: build/${{ matrix.version }}/report.json
 
+      # Build and link non-matching files to verify lack of errors
+      - name: Build non-matching
+        run: |
+          python configure.py --non-matching --map \
+              --version ${{ matrix.version }} \
+              --binutils /binutils --compilers /compilers
+          ninja all_source
+
   website:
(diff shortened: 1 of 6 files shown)


.github/workflows/build.yml
```diff
@@ -73,6 +73,14 @@ jobs:
           name: ${{ matrix.version }}_previous
           path: build/${{ matrix.version }}/report.json
 
+      # Build and link non-matching files to verify lack of errors
+      - name: Build non-matching
+        run: |
+          python configure.py --non-matching --map \
```
SquareMan: I don't thin we need `--map` here. It slows down the build a lot. I think we only use it above for some sort of progress/metrics calculation.

SquareMan (changes requested):
Looking at the output I'm pretty sure it's not actually building. We may have to delete the existing binary to get it to rebuild.

SquareMan:
CI looks good now. Do you think you'd be able to fix the linker error now? Some data just needs to be moved from a header to the proper cpp file. I can also go ahead and merge anyway and fix it in a followup.

JoshSanch:
I'd like to fix it before we merge - currently working on fixing the linker errors.

PR #483, opened 2025-03-28
CI: Split build action into three parallel actions

JoshSanch:


.github/workflows/build.yml
```diff
@@ -67,10 +59,76 @@ jobs:
             build/${{ matrix.version }}/report.json
             build/${{ matrix.version }}/progress-commit.json
 
+  verifyNonMatching:
```
encounter:
```suggestion
  build-non-matching:
```
For consistency with `build` and avoiding camelCase

.github/workflows/build.yml
```diff
@@ -67,10 +59,76 @@ jobs:
...
+      matrix:
+        # Add game versions here
+        version: [GQPE78]
+
+    steps:
+      # Checkout the repository
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 2
```
encounter:
```suggestion
```
Only the previous report job needs this (so it can be removed from the main build job too)

.github/workflows/build.yml
```diff
@@ -67,10 +59,76 @@ jobs:
...
+
+      # Build and link non-matching files to verify lack of errors
+      - name: Build non-matching
+        run: |
+          python configure.py --non-matching \
+              --version ${{ matrix.version }} \
+              --binutils /binutils --compilers /compilers
+          ninja
+
+  generatePreviousReport:
```
encounter:
```suggestion
  build-previous:
```
Same as above

.github/workflows/build.yml
```diff
@@ -67,10 +59,76 @@ jobs:
...
+      - name: Prepare
+        run: cp -R /orig .
+
       # Generate report for previous commit
       - name: Generate previous report
         continue-on-error: true
         run: |
+          python configure.py --version ${{ matrix.version }} \
+                --binutils /binutils --compilers /compilers
           git checkout HEAD^
```
encounter:
```suggestion
          git checkout HEAD^
          python configure.py --version ${{ matrix.version }} \
                --binutils /binutils --compilers /compilers
```
Reorder the operations here

encounter (commented):
Let's give every job a `name:`, so that they show up nicer in the Actions tab. Suggestions:
- Build matching
- Build non-matching
- Build previous

.github/workflows/build.yml
```diff
@@ -85,7 +143,10 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: write
-    needs: build
+    needs: 
+      - build
+      - verifyNonMatching
+      - generatePreviousReport
```
encounter:
```suggestion
      - build
      - build-previous
```
website doesn't technically rely on anything from the non-matching build

encounter (approved):
Looks great!

PR #485, opened 2025-03-29
More zNPCMgr 100% matches

joeyballentine:


src/SB/Game/zNPCMessenger.h
```diff
@@ -10,23 +10,26 @@ void zNPCMsg_Startup();
...
                         en_NPCTYPES toNPCType);
+void zNPCMsg_Timestep(xScene* xscn, F32 dt);
 
-struct NPCPSClt {
+struct NPCPSClt
+{
     // total size: 0x10
-    void (* notify)(void *, class NPCMsg *); // offset 0x0, size 0x4
-    void * notedata; // offset 0x4, size 0x4
+    void (*notify)(void*, class NPCMsg*); // offset 0x0, size 0x4
```
JoshSanch: Class keyword is unnecessary artifact, I'm assuming from DWARF data. Please remove

src/SB/Game/zNPCMessenger.h
```diff
@@ -10,23 +10,26 @@ void zNPCMsg_Startup();
...
+    void* notedata; // offset 0x4, size 0x4
     signed int flg_filter; // offset 0x8, size 0x4
-    class NPCPSClt * next; // offset 0xC, size 0x4
+    class NPCPSClt* next; // offset 0xC, size 0x4
```
JoshSanch: Remove class keyword here as well

JoshSanch (changes requested):

joeyballentine:
Decomped another function. While doing this I noticed we had an enum not filled out. The dwarf data had it filled out, but its values were ints, so i wrote a script to convert those to the correct char codes.

PR #487, opened 2025-03-29
Improve zScene matchingness

escape209:
These string definitions aren't needed anymore. Removing them allows several functions to 100% from correct stringBase offsets. Also get zSceneSetOldScreenAdj match with hack.

### DIFF
--- a/src/SB/Game/zScene.cpp
+++ b/src/SB/Game/zScene.cpp
@@ -110,8 +110,8 @@ static S32 neededFiles;
 static F32 offsetx;
 static F32 offsety;
 static U32 enableScreenAdj;
-static F32 oldOffsetx;
-static F32 oldOffsety;
+volatile static F32 oldOffsetx;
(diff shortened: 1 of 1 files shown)


src/SB/Game/zScene.cpp
```diff
@@ -3512,8 +3498,8 @@ void zSceneEnableScreenAdj(U32 enable)
 
 void zSceneSetOldScreenAdj()
 {
-    oldOffsetx = offsetx;
-    oldOffsety = offsety;
+    *(volatile F32*)(&oldOffsetx) = offsetx;
+    *(volatile F32*)(&oldOffsety) = offsety;
```
JoshSanch: Rather than casting the variables here like this, it would be preferable to leave the simpler code intact and add `volatile` to the "old" variable declarations. From what I can see, no other code makes use of these values.

JoshSanch (changes requested):

PR #492, opened 2025-03-30
zNPCTiki_InitStacking

joeyballentine:
Won't be 100% until FindParents is decomped.


src/SB/Game/zNPCTypeTiki.cpp
```diff
@@ -21,6 +21,8 @@ extern F32 _867;
 extern NPCSndTrax g_sndTrax_TikiShared[3];
 extern NPCSndTrax g_sndTrax_TikiThunder[2];
 
+static class zNPCTiki* orphanList;
```
JoshSanch: Remove class keyword

PR #495, opened 2025-03-31
Kingjelly & prawn cleanup, some kj functions

LivewireCB:


src/SB/Game/zNPCTypePrawn.cpp
```diff
@@ -233,34 +233,23 @@ void tweak_group::register_tweaks(bool init, xModelAssetParam* ap, U32 apsize)
...
-    //   return;
 }
 
 void aqua_beam::start() //100% code match, data does not match
 {
     firing = 1;
-    time = 0;
-    ring.emit_time = 0;
+    time = 0.0;
+    ring.emit_time = -1000000000.0;
```
JoshSanch: Are these meant to be floats or doubles? I haven't seen any examples of doubles in the code so if these are meant to be floats, you need to append the 'f' specifier.

PR #504, opened 2025-04-02
CI: Use pull_request event payload for pull_request_target events

JoshSanch:


mkst:
Probably a known thing, but due to `pull_request_target`, the checks for this PR use the yaml from the main branch (rather than from this PR)

JoshSanch:
Oh shoot. That might actually just do it. I'll revert the string wrap commit and just YOLO merge, then make a subsequent test change PR to verify. Thanks!

PR #511, opened 2025-04-05
xMath3 matches

escape209:
(Forgot to remove iMath3 include but we'll need it later anyway)

Intrinsics my beloved


SquareMan:
We have a macro `xAbs` btw for `(F32)__fabs`. Just for future reference.

PR #518, opened 2025-04-14
40% - 92% for some NCIN_Par functions in zNPCFXCinematic

Redberd36:


src/SB/Game/zNPCFXCinematic.cpp
```diff
@@ -1,13 +1,18 @@
...
 #include "zNPCFXCinematic.h"
 
 #include <types.h>
-#include "xCutscene.h"
 
 static S32 init = 0;
 
+static const F32 zero = 0.0f;            
+static const F32 addVal = 20.0f;         
+static const F32 fillVec3Val = 5.0f;
```
JoshSanch: These symbols do not exist in the symbol map contained within objdiff. You'll need to get rid of them and replace them with their literal values where used to increase the matchingness.

src/SB/Game/zNPCFXCinematic.cpp
```diff
@@ -20,6 +25,547 @@ void zNPCFXCinematic_Init()
...
+        return;
+    }
+
+    if (fxrec->flg_stat & 2)
+    {
+        fxrec->fxdata.pardata.emitter = zParEmitterFind("PAREMIT_BPLANK_JET_1");
+
+        if (fxrec->fxdata.pardata.emitter->tasset)
+        {
+            if (((U8*)fxrec->fxdata.pardata.emitter->tasset)[9] != 15)
```
JoshSanch: This doesn't seem like what was actually written here. See if you can figure out what the equivalent (fxrec->fxdata.pardata.emitter->tasset + 0x9) should be when accessing the data in the tasset.

src/SB/Game/zNPCFXCinematic.cpp
```diff
@@ -20,6 +25,547 @@ void zNPCFXCinematic_Init()
...
+        }
+    }
+
+    if (fxrec->fxdata.pardata.emitter == NULL)
+    {
+        fxrec->flg_stat |= 4;
+        return;
+    }
+
+    F32 & float_accum = *(F32*)((U8*)fxrec + 40);
```
JoshSanch: Try to write this in terms of a struct member and not an indefinite char offset from the fxrec pointer. It should look like your other code (e.g. fxrec->flg_stat, fxrec->fxdata).

JoshSanch (changes requested):
Good work nailing a lot of progress!

Please update this PR to the best of your ability to address feedback. In particular, there's a lot of pointer magic happening for struct member accesses. The DWARF data should provide you everything you need to translate things like:
```
        U8 val = ((U8*)fxrec->fxdata.pardata.emitter)[18];
```
to proper struct member accesses.

src/SB/Game/zNPCFXCinematic.cpp
JoshSanch: References for this statement:
image
image

src/SB/Game/zNPCFXCinematic.cpp
Redberd36: Took some digging - but I found it!

"fxrec->fxdata.pardata.emitter->tasset->emit_type"

PR #519, opened 2025-04-17
Working more on zNPCFXCinematic with one 100% and started on other functions

Redberd36:


src/SB/Game/zNPCFXCinematic.cpp
```diff
@@ -460,7 +460,7 @@ void NCIN_BubSlam(const zCutsceneMgr*, NCINEntry* fxrec, S32 param)
     return;
     }
 
-    switch (fxrec->typ_ncinfx)
+   switch(fxrec->typ_ncinfx)
```
mattbruv: Can you run the formatter on this file so things like this don't sneak into PRs?

src/SB/Game/zTalkBox.cpp
```diff
@@ -13,6 +13,22 @@ namespace
...
+    *sharedMemory = value;
+    *(sharedMemory + 8) = value;
+    *(sharedMemory + 9) = 0;
+    *(sharedMemory + 10) = 0;
+    *(sharedMemory + 11) = 0;
+    *(sharedMemory + 12) = 0;
+
+    clear_layout(); //FIXME: Find and replace with xtextbox::layout::clear()
+    return;
+}
```
mattbruv: This is something I wouldn't want to include, it's very rough and only raises the function to 28%, so it's a big work in progress still, and doesn't necessarily give us anything useful to work off of. I'd remove it for now from this PR.

mattbruv (changes requested):

PR #539, opened 2025-05-12
Match zCamera sdata2 section

stravant:
* Migrate extern floats to actual constants.

* Tweak the existing constants to be bitwise correct.

* Move some forward declares to the anonymous namespace they should have been in.

* Start reversing zCameraUpdate.

### DIFF
--- a/src/SB/Game/zCamera.cpp
+++ b/src/SB/Game/zCamera.cpp
@@ -97,28 +97,12 @@ extern zGlobals globals;
 extern const xVec3 g_O3;
 extern F32 gSkipTimeFlythrough;
 
-extern F32 zCamera_f_75_0; // 75.0
-extern F32 zCamera_f_0_0; // 0.0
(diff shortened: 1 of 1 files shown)


stravant:
(Also, didn't mean to branch on the main repo instead of the fork, used to doing that from another project at the moment so muscle memory)

src/SB/Game/zCamera.cpp
```diff
@@ -890,7 +897,7 @@ void zCameraMinTargetHeightSet(F32 min_height)
 
 void zCameraMinTargetHeightClear()
 {
-    zcam_mintgtheight = zCamera_f_n1_0_e38;
+    zcam_mintgtheight = -1.0e38;
```
JoshSanch: Missed an f suffix here, would be good to add to avoid accidental cast to float and to keep the style consistent.

PR #542, opened 2025-05-15
xPtankPool: implement several ptank functions

velkog:
Implement sort_buckets, create_ptank, destroy_ptank, render_ptank, compare_ptanks, andxPTankPoolRender

All fully matching except for create_ptank, which is having mismatches with float assignments.

### DIFF
--- a/src/SB/Core/x/xPtankPool.cpp
+++ b/src/SB/Core/x/xPtankPool.cpp
@@ -1,3 +1,262 @@
 #include "xPtankPool.h"
 
+#include <rpptank.h>
+#include <stdlib.h>
+#include <string.h>
 #include <types.h>
+
+extern RwInt32 _rpPTankAtomicDataOffset;
+
+namespace
+{
+
+#define NUM_GROUPS 2
+#define MAX_PARTICLES 64
+
+struct group_data
+{
+    U32 max_size;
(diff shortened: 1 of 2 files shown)


src/SB/Core/x/xPtankPool.cpp
```diff
@@ -1,3 +1,262 @@
...
+    if (!this->valid())
+    {
+        goto cleanup;
+    }
+
+    RpPTankAtomicUnlock(this->ptank);
+
+cleanup:
+    return;
+}
```
JoshSanch: Does the goto here help with matchingness? I would imagine substituting the goto with a return would be equivalent but I can't verify that atm.

velkog: Oh woops. This function is actually still very incomplete (only 6% match). This was just the start of my efforts and then I got distracted and forgot about it haha. I think a return doesn't work because there is more that is needed after the cleanup portion, but tbd on further implementation. I can remove this from the PR if we want this cleaned up.

JoshSanch: Yeah, I think this is best left omitted for now. No worries!

velkog: Removed!

PR #543, opened 2025-05-15
xCutscene: Partial/Full matches, cutsceneHackTable update

energydrink02:
Implement xCutscene_Update, CutsceneShadowRender and xCutscene_SetSpeed. Fully match except for SetSpeed (register swap).

Update cutsceneHackTable in zCutsceneMgr


src/SB/Core/x/xCutscene.cpp
```diff
@@ -148,6 +148,105 @@ S32 xCutscene_LoadStart(xCutscene* csn)
...
+    {
+        if (csn->SndNumChannel == 1)
+        {
+            xSndSetPitch(csn->SndHandle[0], 0.0f);
+        }
+        else if (csn->SndNumChannel == 2)
+        {
+            xSndStartStereo(csn->SndHandle[0], csn->SndHandle[1], 0.0f);
+        }
+        csn->SndStarted = 1;
```
JoshSanch: Opt for using the TRUE and FALSE macros for 0/1 when more semantically appropriate.

In this case, I would recommend doing `csn->SndStarted == FALSE` and `csn->SndStarted == TRUE`. Anything being used as a C-style boolean should use these when appropriate.

SndNumChannel and SndHandle aren't like that so just for cases where SndStarted is used would be enough.

src/SB/Core/x/xCutscene.cpp
```diff
@@ -148,6 +148,105 @@ S32 xCutscene_LoadStart(xCutscene* csn)
...
+        {
+            return 1;
+        }
+
+        if (csn->Waiting != 0)
+        {
+            csn->Time = csn->Play->EndTime;
+            csn->CamTime = xCutsceneConvertBreak(csn->Time, csn->BreakList, csn->Info->BreakCount, -1);
+
+            if (csn->BadReadPause == 0)
```
JoshSanch: This also seems like a C-Style boolean which could benefit from TRUE/FALSE.

src/SB/Core/x/xCutscene.cpp
energydrink02: done

src/SB/Core/x/xCutscene.cpp
energydrink02: done

JoshSanch (approved):
Looks good to me. Thanks for your first contribution!

PR #551, opened 2025-05-17
iTRC work

LivewireCB:
iTRC work and some file cleanup

### DIFF
--- a/include/dolphin/gx/GXGeometry.h
+++ b/include/dolphin/gx/GXGeometry.h
@@ -36,6 +36,22 @@ static inline void GXEnd(void) {
 #endif
 }
 
+static inline void GXTexCoord2s16(int r3, int r4) // possible inline req.
+{
+    int ptr = 0xcc010000;
+    *(short*)((char*)(ptr) - 0x8000) = r3;
+    *(short*)((char*)(ptr) - 0x8000) = r4;
+}
+
+
+static inline void GXPosition3s16(int r3, int r4, int r5) // possible inline req.
+{
+    int ptr = 0xcc010000;
+    *(short*)((char*)(ptr) - 0x8000) = r3;
+    *(short*)((char*)(ptr) - 0x8000) = r4;
+    *(short*)((char*)(ptr) - 0x8000) = r5;
+}
+
 void GXSetLineWidth(u8 width, GXTexOffset texOffsets);
 void GXSetPointSize(u8 pointSize, GXTexOffset texOffsets);
 void GXEnableTexOffsets(GXTexCoordID coord, u8 line_enable, u8 point_enable);
--- a/src/SB/Core/gc/iTRC.cpp
+++ b/src/SB/Core/gc/iTRC.cpp
@@ -69,36 +69,36 @@ BOOL ROMFont::Init()
 
 void ROMFont::InitDisplay(_GXRenderModeObj* InRenderMode)
 {
-  GXColor clr = { 0 };
-  Mtx44 mtx;
-  Mtx idt;
-  
-  mRenderMode = InRenderMode;
-  void** xfb = &mXFBs[1];
-  RwGameCubeGetXFBs(&mXFBs[0], &mXFBs[1]);
-  mCurrentFrameBuffer = *xfb;
+    GXColor clr = { 0 };
+    Mtx44 mtx;
+    Mtx idt;
 
-  ROMFont::InitGX();
-  ROMFont::InitVI();
-  
-  GXSetCopyClear(clr,0xffffff);
-  C_MTXOrtho(mtx, 0.0f, mRenderMode->efbHeight, 0.0f, mRenderMode->fbWidth, 0.0f, 10000.0f);
-  GXSetProjection(mtx, GX_ORTHOGRAPHIC);
-  PSMTXIdentity(idt);
-  GXLoadPosMtxImm(idt, 0);
-  GXSetCurrentMtx(0);
-  GXSetZMode(TRUE, GX_ALWAYS, TRUE);
-  GXSetNumChans(0);
-  GXSetNumTevStages(1);
-  GXSetTevOp(GX_TEVSTAGE0, GX_REPLACE);
-  GXSetTevOrder(GX_TEVSTAGE0, GX_TEXCOORD0, GX_TEXMAP0, GX_COLOR_NULL);
-  GXSetBlendMode(GX_BM_BLEND, GX_BL_ONE, GX_BL_ONE, GX_LO_CLEAR);
-  GXClearVtxDesc();
-  GXSetVtxDesc(GX_VA_POS, GX_DIRECT);
-  GXSetVtxDesc(GX_VA_TEX0, GX_DIRECT);
-  GXSetVtxAttrFmt(GX_VTXFMT0, GX_VA_POS, GX_NRM_NBT, GX_RGBA4, 0);
-  GXSetVtxAttrFmt(GX_VTXFMT0, GX_VA_TEX0, GX_NRM_NBT, GX_RGBA4, 0);
-  return;
+    mRenderMode = InRenderMode;
+    void** xfb = &mXFBs[1];
+    RwGameCubeGetXFBs(&mXFBs[0], &mXFBs[1]);
+    mCurrentFrameBuffer = *xfb;
+
+    ROMFont::InitGX();
+    ROMFont::InitVI();
+
+    GXSetCopyClear(clr, 0xffffff);
+    C_MTXOrtho(mtx, 0.0f, mRenderMode->efbHeight, 0.0f, mRenderMode->fbWidth, 0.0f, 10000.0f);
+    GXSetProjection(mtx, GX_ORTHOGRAPHIC);
+    PSMTXIdentity(idt);
+    GXLoadPosMtxImm(idt, 0);
+    GXSetCurrentMtx(0);
+    GXSetZMode(TRUE, GX_ALWAYS, TRUE);
+    GXSetNumChans(0);
+    GXSetNumTevStages(1);
+    GXSetTevOp(GX_TEVSTAGE0, GX_REPLACE);
+    GXSetTevOrder(GX_TEVSTAGE0, GX_TEXCOORD0, GX_TEXMAP0, GX_COLOR_NULL);
+    GXSetBlendMode(GX_BM_BLEND, GX_BL_ONE, GX_BL_ONE, GX_LO_CLEAR);
+    GXClearVtxDesc();
+    GXSetVtxDesc(GX_VA_POS, GX_DIRECT);
+    GXSetVtxDesc(GX_VA_TEX0, GX_DIRECT);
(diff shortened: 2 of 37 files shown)


src/SB/Core/gc/iTRC.cpp
```diff
@@ -289,11 +331,49 @@ void ResetButton::SetSndKillFunction(void (*Func)())
...
+    }
+    else {
+      OSResetSystem(0,0,0); // bool = osresetsystem
+      uVar2 = (U32)tempblank;
+      uVar1 = mWasResetButtonPressed;
+    }
+  }
+  mWasResetButtonPressed = uVar1;
+  //return uVar2;
 }
```
JoshSanch: Good stuff overall! Please clean up some of the Ghidraisms here and use bool macros for the appropriate conditional checks.

For example, (code*) 0x0 can just be rewritten as NULL, and the conditionals that are checking 0 values are replaceable with the more semantic FALSE.

We also avoid using the comma operator in conditionals because it makes the conditions harder to read - please convert those to statements outside the condition using a nested conditional.

JoshSanch (changes requested):

LivewireCB:
Lots of work to zMain (mostly done manually). Out of time to finish. Will resume later.

include/dolphin/gx/GXGeometry.h
```diff
@@ -36,6 +36,22 @@ static inline void GXEnd(void) {
...
+
+
+static inline void GXPosition3s16(int r3, int r4, int r5) // possible inline req.
+{
+    int ptr = 0xcc010000;
+    *(short*)((char*)(ptr) - 0x8000) = r3;
+    *(short*)((char*)(ptr) - 0x8000) = r4;
+    *(short*)((char*)(ptr) - 0x8000) = r5;
+}
+
```
JoshSanch, line 54: These should get the Heavy Iron datatype treatment - U32, U16*. `char*` is fine as it's sometimes used intentionally for string ops.

Also, are these subtraction operations in the ASM or is it possible to change this to a constant value `0xCC008000`?

src/SB/Core/gc/iTRC.cpp
```diff
@@ -289,11 +331,20 @@ void ResetButton::SetSndKillFunction(void (*Func)())
 
 void ResetButton::CheckResetButton()
 {
+// TODO
+
+  OSGetResetButtonState();
+  PADRecalibrate(0xf0000000);
+  VISetBlack(1);
```
JoshSanch: This spot would benefit from the TRUE macro.

src/SB/Core/gc/iTRC.cpp
```diff
@@ -289,11 +331,20 @@ void ResetButton::SetSndKillFunction(void (*Func)())
...
+  VIWaitForRetrace();
+  OSEnableInterrupts();
+  DVDGetDriveStatus();
 }
 
-bool iTRCDisk::Init(void)
+void iTRCDisk::Init()
 {
-    return ROMFont::Init();
+    ROMFont::Init();
```
JoshSanch, line 346: Could you share the concrete reasoning behind this signature change? Would like to document instances like this where we get it wrong on initial definition to aid future efforts.

src/SB/Core/gc/iTRC.cpp
```diff
@@ -338,11 +389,42 @@ void iTRCDisk::SetMovieResumeFunction(void (*Func)())
...
 {
-    return false;
+  S32 diskChk;
+  bool diskIDed;
+  
+  diskIDed = 0;
+  diskChk = DVDCheckDisk();
+
+  if ((diskChk != 0) && (diskChk = DVDGetDriveStatus(), diskChk == 0)) {
+    diskIDed = 1;
```
JoshSanch: Please use for TRUE/FALSE macros for diskChk conditions.

Also, the conditional body should read as this instead since `diskIDed` is a boolean:
```cpp
diskIDed = true;
```

src/SB/Core/gc/iTRC.cpp
```diff
@@ -338,11 +389,42 @@ void iTRCDisk::SetMovieResumeFunction(void (*Func)())
...
+  }
+  return diskIDed;
 }
 
 void iTRCDisk::DisplayErrorMessage()
 {
+  S32 tempVar;
+  U32 local_18 [4];
```
JoshSanch: This variable isn't used in the function, so it should be removed.

src/SB/Core/x/xCamera.h
```diff
@@ -10,6 +10,64 @@
...
+    xVec3 epicenter;
+    xVec3* player;
+};
+
+struct cameraFXZoom
+{
+    F32 holdTime;
+    F32 vel;
+    F32 accel;
+    F32 distance;
```
JoshSanch, line 33: Thanks for moving these :)

src/SB/Game/zMain.cpp
```diff
@@ -25,6 +103,414 @@ void zMainInitGlobals()
...
+    globals.player.g.BBashTime = xIniGetFloat(ini,"g.BBashTime", 0.3f); // @1005;
+    globals.player.g.BBashDelay = xIniGetFloat(ini,"g.BBashDelay", 0.15f); // @1006;
+    globals.player.g.BBashCVTime = xIniGetFloat(ini,"g.BBashCVTime", 0.15f); // @1006;
+    globals.player.g.BBashHeight = xIniGetFloat(ini,"g.BBashHeight", 3.0f); // @1007;
+    globals.player.g.BBounceSpeed = xIniGetFloat(ini,"g.BBounceSpeed", 10.0f); // @1008;
+    globals.player.g.BSpinMinFrame = xIniGetFloat(ini,"g.BSpinMinFrame", 2.0f); // @1009;
+    globals.player.g.BSpinMaxFrame = xIniGetFloat(ini,"g.BSpinMaxFrame", 20.0f); // @1010;
+    globals.player.g.BSpinRadius = xIniGetFloat(ini,"g.BSpinRadius", 0.3f); // @1005;
+    globals.player.g.BSpinMinFrame = globals.player.g.BSpinMinFrame * 0.033333335f; // @909
+    globals.player.g.BSpinMaxFrame = globals.player.g.BSpinMaxFrame * 0.033333335f; // @909
```
JoshSanch:
```suggestion
    globals.player.g.BSpinMinFrame *= 0.033333335f; // @909
    globals.player.g.BSpinMaxFrame *= 0.033333335f; // @909
```

src/SB/Game/zMain.cpp
```diff
@@ -25,6 +103,414 @@ void zMainInitGlobals()
...
+    globals.player.g.SandyMeleeMinFrame = xIniGetFloat(ini,"g.SandyMeleeMinFrame", 1.0f); // @1002;
+    globals.player.g.SandyMeleeMaxFrame = xIniGetFloat(ini,"g.SandyMeleeMaxFrame", 9.0f); // @1011;
+    globals.player.g.SandyMeleeRadius = xIniGetFloat(ini,"g.SandyMeleeRadius", 0.3f); // @1005;
+    globals.player.g.SandyMeleeMinFrame = globals.player.g.SandyMeleeMinFrame * 0.033333335f; // @909
+    globals.player.g.SandyMeleeMaxFrame = globals.player.g.SandyMeleeMaxFrame * 0.033333335f; // @909
```
JoshSanch:
```suggestion
    globals.player.g.SandyMeleeMinFrame *= 0.033333335f; // @909
    globals.player.g.SandyMeleeMaxFrame *= 0.033333335f; // @909
```

src/SB/Game/zMain.h
```diff
@@ -1,4 +1,34 @@
...
+
+
+enum eStartupErrors {
+    eNoError = 0,
+    eNoFormat = 1,
+    eDamagedCard = 2,
+    eWrongDevice = 3,
+    eNoCards = 4,
+    eCorruptFile = 5,
+    eNoController = 6,
```
JoshSanch: Since the default mode for enums is starting at value 0, it's more concise to define an enum like this by omitting the explicit values. Most IDEs will show you the integer value if you hover over the enum value. I'll make note of this one for some style guide discussion with the maintenance team. We can leave it **as is**, for now.

To clarify, writing it way does make it easier to decompile usages of the enum, which is a certain benefit of writing it this way.
However, most of our tooling already handles mapping from symbol name to value (Ghidra, m2c), and the values clutter things visually for later down the line when folks will be modifying or otherwise editing these values.

JoshSanch (changes requested):
Excellent contribution! Since this is a big one, there's a few review items I have left feedback on. Additionally, please format all the files in this PR with clang-format - there were at least a couple which were not formatted as such. This will appear in our contribution guidelines going forward.

include/dolphin/gx/GXGeometry.h
LivewireCB: Even though these are SDK files they'd get the heavy iron treatment?
Also, I originally copied this format from another file, but the ASM is:
lis r5, 0xcc01
sth r3, -0x8000(r5)
sth r4, -80x8000(r5)

src/SB/Core/gc/iTRC.cpp
LivewireCB: returning the function created extra ASM instructions that were not supposed to be there. Therefore, I swapped it to a void and took away the return then the function matched 100%

src/SB/Core/gc/iTRC.cpp
LivewireCB: Which one? iirc the variable local_18 is in the dwarf so it theoretically should be used. Since it isn't currently being used I will remove it.

src/SB/Game/zMain.cpp
LivewireCB: Genuinely did not know this existed.

include/dolphin/gx/GXGeometry.h
JoshSanch: Oh you're so right, my bad - didn't check the filename. This seems appropriate, then.

src/SB/Core/gc/iTRC.cpp
JoshSanch: Sounds good - resolving.

src/SB/Core/gc/iTRC.cpp
JoshSanch: What's causing this file not to match? If the unused array is going to end up being required for the 100% match, then I'd leave it in.

src/SB/Core/gc/iTRC.cpp
```diff
@@ -389,41 +388,44 @@ void iTRCDisk::SetMovieResumeFunction(void (*Func)())
...
 {
-  S32 diskChk;
-  bool diskIDed;
-  
-  diskIDed = 0;
-  diskChk = DVDCheckDisk();
+    S32 diskChk;
+    bool diskIDed;
+
+    diskIDed = FALSE;
```
JoshSanch: This is kind of confusing so I totally understand why you used FALSE here, and it technically compiles but:

Use the C++ keyword `false` for working with C++ `bool` data types.
Use the C-style macro FALSE for working with C types (S32, S8, U8, etc.)

Same applies for TRUE/true usage.

JoshSanch (changes requested):

PR #552, opened 2025-05-18
zCameraTweak: Progress

0x5abe:
I don't really know if it's cause of float scheduling or if there are any mistakes in the implementations, but zCameraTweakGlobal_Add, zCameraTweakGlobal_Remove, zCameraTweakGlobal_Reset and zCameraTweakGlobal_Update are really messed up. Also some of the data in sdata2 isn't being generated. In any case, these 4 functions can be removed if the PR can only be merged with the stuff that is fully matching.

### DIFF
--- a/src/SB/Game/zCameraTweak.cpp
+++ b/src/SB/Game/zCameraTweak.cpp
@@ -3,7 +3,6 @@
 #include "xMath.h"
 #include "xMathInlines.h"
 #include "iMath.h"
-
 #include "zCamera.h"
 #include <types.h>
 
@@ -86,6 +85,7 @@ void zCameraTweakGlobal_Add(U32 owner, F32 priority, F32 time, F32 pitch, F32 di
 void zCameraTweakGlobal_Remove(U32 owner)
 {
     S32 i;
+    S32 j;
 
     for (i = 0; i < sCamTweakCount; i++)
     {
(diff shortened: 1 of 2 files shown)


src/SB/Game/zCameraTweak.cpp
```diff
@@ -1,22 +1,229 @@
...
+static F32 sCamPitch;
+static zCamTweakLook zcam_neartweak;
+static zCamTweakLook zcam_fartweak;
 
-F32 zCameraTweakGlobal_GetPitch()
+static void zCameraTweak_LookPreCalc(zCamTweakLook* tlook, F32 d, F32 h, F32 pitch)
 {
-    return sCamPitch;
+    F32 tan = itan(pitch);
+    F32 unk = d * tan;
```
JoshSanch: This operation is a [trig function](https://www.cuemath.com/trigonometry/tangent-function/) meant to calculate a vertical offset given an angle and an adjacent side d, which we can presume to be distance based on the symbol name.

A better name for unk, therefore, would be deltaH, since the code seems to be modifying the original height by this incremental value.

JoshSanch (changes requested):

src/SB/Game/zCameraTweak.cpp
0x5abe: Just updated this! Thanks for the explanation

PR #553, opened 2025-05-18
Start zShrapnel

energydrink02:
Started working on zShrapnel with some full matched functions and several equivalents (left some functions that are beyond my capabilities empty but needed for matching other functions/variables)

### DIFF
--- a/src/SB/Game/zShrapnel.cpp
+++ b/src/SB/Game/zShrapnel.cpp
@@ -1,3 +1,890 @@
+#include <types.h>
+#include "iModel.h"
+
+#include "xColor.h"
+#include "xSnd.h"
+#include "xstransvc.h"
+
+#include "zGlobals.h"
 #include "zShrapnel.h"
+#include "zLightning.h"
+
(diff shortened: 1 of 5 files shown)


src/SB/Game/zShrapnel.cpp
```diff
@@ -1,3 +1,890 @@
...
+        curr++;
+    }
+}
+
 
 #include <types.h>
+
+void zShrapnel_Update(F32 dt)
+{
+    if (sNumActiveFrags == 0)
```
mattbruv: Probably unintentional, but let's try to keep `#include` statements at the top of the files for clarity sake

mattbruv (changes requested):
Excellent work, just address the one issue and this should be good to merge.

src/SB/Game/zShrapnel.cpp
energydrink02: Whops, yea blame vscode

mattbruv (approved):

PR #554, opened 2025-05-19
Docs: First draft of contribution guidelines

JoshSanch:
With all of the new contributors coming in lately, I'd like to take this moment to introduce some contribution guidelines to help folks make PRs that will be easy to review and merge into the project. 

My goal with this document is to have an accessible list of expectations that contributors can reference ahead of time and allowing for in-depth conversations on guidelines to occur within pull requests when they infrequently occur.

\- Josh


LivewireCB:
Consider adding a clear definition of variable naming conventions.
for example; using "myVar" instead of "MyVar" or "myvar"

CONTRIBUTING.md
```diff
@@ -0,0 +1,29 @@
...
+- A good-faith attempt to clean up code generated by decompiler tools (m2c, Ghidra) is expected, using DWARF data and project code conventions as appropriate.
+- Encourage and support other contributors, especially new ones! Decompiling games is hard and difficult work, but sharing knowledge and supporting fellow decompers makes that work a lot easier and more enjoyable for all of us.
+
+## Code Conventions
+- Avoid using pointer math for unknown struct members. Instead, populate the struct with padding and the appropriate type using `unk_<offset>` as the naming convention. For example, define an unknown float member at offset 0xF0 as `F32 unk_F0`.
+- Use the `TRUE`/`FALSE` macros in contexts where conditions use the value of a non-bool data type (S32, U8) as a boolean-like.
+- Use the `NULL` macro when assigning `0`/`0x0` to a pointer value.
+- Do not explicitly define enum values, unless the starting value is non-zero or the increment is non-standard.
+- Use PowerPC naming conventions to provide variable names for values which have no symbols contained in the available DWARF data, i.e:
```
SquareMan: I think I'd prefer a bit more nuance here. I personally try to reason about what the symbol actually means/does before falling back to a placeholder name

also in sufficiently complex functions, registers may be reused if the variable's lifetimes don't overlap.

Otherwise I think I'm fine with this convention, so I'd just phrase it as a first choice for a placeholder name or something like that.

JoshSanch: I support that premise. Maybe something like..?
```suggestion
- Use PowerPC naming conventions when creating placeholder variable names. Placeholders should only be used when no DWARF symbol data is available and the variable does not have a name that can be reasonably assigned based on its usage.
```

Additionally, I feel that especially with complex functions, even if there is overlap and the variable is used multiple times, it makes the code a little easier to decompile since it reads closer to its assembly equivalent. I'm open to discussion on that point, though.

SquareMan: I like that change.

My point on the reuse thing was more that you may unable to use the same the name for both, for example if they were different types and in the same lexical scope. I don't know if this case is really very common though.

JoshSanch: That's a good point, the type might vary from usage to usage. Maybe something like r3_2 or r3_3 for subsequent declarations? Although it starts to look a little bit ugly - but like you mentioned this case isn't very common so we're really discussing naming theory for general PowerPC decomp projects. I'm not aware of a standard.

JoshSanch:
Candidate convention I'd like to add based on [this commit](https://github.com/bfbbdecomp/bfbb/pull/561/commits/3166a819cf89caa720af1c5e1e1092ec04d918ae):
```
- Hexadecimal literals used as arguments for asset functions should be represented using char literals when it would improve the readability of the code (i.e 0x504c4154 -> 'PLAT').
```

JoshSanch:
I'm adding this line - it's a bit arbitrary but it closely aligns with Heavy Iron's naming standards so I'd like to follow that:

```
- When DWARF symbols are unavailable, ensure variables are named using camelCase and structs/classes are named using TitleCase, or snake_case if they exist only in an anonymous namespace.
```

CONTRIBUTING.md
SquareMan: That's fine with me.

CONTRIBUTING.md
```diff
@@ -0,0 +1,32 @@
...
+
+This documentation outlines our project contribution guidelines. Following these guidelines allows the maintainers review changes faster and give better feedback, and helps to ensure that future modding and PC port efforts are made as simple as possible.
+
+In addition to decompilation efforts, this project also welcomes tooling and [tutorial/knowledge resource](https://github.com/bfbbdecomp/bfbb-decomp-guide) contributions! Check out [our organization page](https://github.com/bfbbdecomp) to see all of the projects we welcome contributions to.
+
+## Contribution Guidelines
+
+Please respect the following guidelines:
+- Ensure your code is formatted using the project's clang-format autoconfiguration ([VS Code Plugin](https://marketplace.visualstudio.com/items?itemName=xaver.clang-format)).
+- Keep pull requests scoped to a single compilation unit when possible. Defining functions in other files is encouraged, however, in cases where the function is used by the compilation unit you're contributing to.
```
SquareMan: This second sentence reads a little clunky to me with the "however" in the middle. This feels clearer to me:
> However, defining functions in other files is encouraged in cases where the function is used by the compilation unit you're contributing to.

SquareMan (approved):
Mostly lgtm.

Feel free to ignore my last comment if you disagree.

CONTRIBUTING.md
JoshSanch: There's probably a reason that I agree with you here but I haven't studied language enough to know why haha. I'll swap it out for your suggestion.

JoshSanch: After some thought, I think the "However" needs to be at the beginning of the follow-up sentence in order to immediately draw attention to the addendum.

JoshSanch:
Proposing to add:
```
- Avoid adding functions or data definitions that will need to be cleaned up and removed later for the sake of increasing matchingness in the short-term.
```

This is inspired by the existing [_all_the_floats](https://github.com/bfbbdecomp/bfbb/blob/8bb2c432c141b5120ea0d0c764b1ded2dcb50a06/src/SB/Game/zEntPlayerBungeeState.cpp#L33) function, and [proposed changes](https://github.com/bfbbdecomp/bfbb/pull/567#discussion_r2103050517) on #567. This is an artifact of the old build/match comparison tooling and is no longer necessary. It also adds clean-up work later for no long-term benefit.

JoshSanch:
Going to merge for now - I invite anyone with any further thoughts to raise an issue or to start discussion in the community Discord.

PR #556, opened 2025-05-19
Partial match for xFuncPiece_EndPoints

daft7:
Partial match for xFuncPiece_EndPoints. xFuncPiece Eval and ShiftPiece still tricky for me to attack.

### DIFF
--- a/src/SB/Core/x/xMath.cpp
+++ b/src/SB/Core/x/xMath.cpp
@@ -436,3 +436,13 @@ void xAccelStop(F32& x, F32& v, F32 a, F32 dt) {
         v = 0.0f;
     }
 }
+
+void xFuncPiece_EndPoints(xFuncPiece* func, F32 pi, F32 pf, F32 fi, F32 ff) {
+    float xfinv;
+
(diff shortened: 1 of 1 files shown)


src/SB/Core/x/xMath.cpp
```diff
@@ -436,3 +436,13 @@ void xAccelStop(F32& x, F32& v, F32 a, F32 dt) {
         v = 0.0f;
     }
 }
+
+void xFuncPiece_EndPoints(xFuncPiece* func, F32 pi, F32 pf, F32 fi, F32 ff) {
+    float xfinv;
```
JoshSanch: Is this from the DWARF data? If so, you can leave it in, but if not, please remove it. 

If you leave it, please change its type to F32.

JoshSanch (changes requested):
Thanks for the contribution! Looking good for the most part, I left some minor feedback. Please also rebase against main and push so that OK Bot can verify that there are no regressions.

JoshSanch (approved):

PR #562, opened 2025-05-21
Finish remainder of xPtankPool

velkog:
This gets all of xPtankPool to match at 100% with the exception of `create_ptank()` which is having scheduling issues with floats.

Changes included are:

- Implement: `init_groups()`, `create_ptanks(...)`, `xPTankPoolSceneEnter()`, `xPTankPoolSceneExit()`, `ptank_pool::grab_block(...)`, and `ptank_pool::flush()`
- Additionally, adjusted some of the function signatures of already matched functions to match what is listed in the PS2 dwarf
- Moved the definition of `ptank_context` which one of the functions excepted to be within the anonymous namespace
- Removed `NUM_GROUPS` in favor of already existing enum value: MAX_PGT
- Updated some hardcoded enum values
- Formatted the entire files `xPtankPool.cpp` and `xPtankPool.h`

image

### DIFF
--- a/src/SB/Core/x/xPtankPool.cpp
+++ b/src/SB/Core/x/xPtankPool.cpp
@@ -75,7 +75,7 @@ namespace
                     {
                         if (it->dst_blend == front->dst_blend &&
                             it->src_blend == front->src_blend &&
-                            (!(it->flags ^ front->flags) & rpPTANKDFLAGPOSITION))
+                            !((it->flags ^ front->flags) & rpPTANKDFLAGPOSITION))
                         {
                             break;
                         }
--- a/src/SB/Core/x/xPtankPool.cpp
+++ b/src/SB/Core/x/xPtankPool.cpp
@@ -14,280 +14,266 @@ namespace
 
 #define MAX_PARTICLES 64
 
-static U8 inited;
+    static U8 inited;
 
-struct ptank_context
-{
-    ptank_context* next;
-    class RpAtomic* ptank;
-    U32 flags;
-    U32 src_blend;
-    U32 dst_blend;
-};
-
-struct group_data
-{
-    U32 max_size;
-    U32 create_flags;
-    ptank_context* ptanks;
-    ptank_context** buckets;
-    U32 size;
-    U32 used;
-    U32 buckets_used;
-};
-
-static group_data groups[MAX_PGT] = {
+    struct ptank_context
     {
-        4, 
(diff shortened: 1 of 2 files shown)


velkog:
Apologies, the changes are a little messy because I didn't run the formatter on the last PR

src/SB/Core/x/xPtankPool.cpp
```diff
@@ -242,8 +343,101 @@ void xPTankPoolRender()
...
+
+                it++;
+            }
+        }
+    }
+}
+
+void ptank_pool::flush()
+{
+    if (this->valid() != 0)
```
JoshSanch: Does this also match? If so, please use the suggestion instead, as it's a bit cleaner.
```suggestion
    if (this->valid())
```

JoshSanch (commented):
Great work overall! I left a comment for discussion - can consider merging after that's addressed.

src/SB/Core/x/xPtankPool.cpp
```diff
@@ -75,7 +75,7 @@ namespace
                     {
                         if (it->dst_blend == front->dst_blend &&
                             it->src_blend == front->src_blend &&
-                            ((it->flags ^ front->flags) & rpPTANKDFLAGPOSITION) == 0)
+                            (!(it->flags ^ front->flags) & rpPTANKDFLAGPOSITION))
```
JoshSanch: It looks like this change has introduced a regression. It should read as follows to be logically equivalent:
```cpp
!((it->flags ^ front->flags) & rpPTANKDFLAGPOSITION)
```

JoshSanch (changes requested):

src/SB/Core/x/xPtankPool.cpp
kvelcich: Got overly confident and didn't check the objdiff

src/SB/Core/x/xPtankPool.cpp
kvelcich: As you probably saw, I also removed all comparisons to zero in the file.

velkog: As you probably saw, I removed a few other direct comparisons to zero in the file

src/SB/Core/x/xPtankPool.cpp
velkog: I got overly confident and didn't check the objdiff before pushing haha

JoshSanch (approved):
Lgtm! Nice work

PR #563, opened 2025-05-21
zNPCTypeKingJelly: register_tweaks work

energydrink02:
Not entirely sure if the namespace stuff is correct — I based it on how it's done in other files.


JoshSanch (changes requested):
Awesome work!

Here, I think it's more accurate to move the function implementations into the namespace, as the tweak_group struct definition here is intended to be local to this file (enforced by the anonymous namespace), and so defining it inside the namespace further reinforces this intention.

JoshSanch (approved):

JoshSanch:
Great work, thanks! :)

PR #564, opened 2025-05-22
zEntPlayerBungeeState: Matches for Several Functions

JoshSanch:
Also includes tweaks to xLightKit and xVec3.


JoshSanch:
Is this the result of my xVec3 changes?? 😵‍💫😵‍💫😵‍💫

stravant:
If you want to include it, put the cross definition in xCollide which is the compilation unit it shows up in, that will fix the regressions.

It's easier to sort out what may or may not be inline later.

mattbruv (approved):

PR #566, opened 2025-05-22
xHud

stravant:
* The annoying one in here is `for_each`. It _almost_ matches minus the handling for the variable size of the elements as a `for (i < count) doStuff(array[i])` style loop, but stops matching well when I try to factor the variable element size in.


JoshSanch (approved):
Lgtm!

PR #567, opened 2025-05-22
xCollide and dependent xMath changes

stravant:
* Progress on xCollide. Stub out many of the methods since I plan to work more on this next.

* Add relevant xMath methods.

* Move some of the math definitions to the xCollide compilation unit they're in to get it close to linking with the decomped version.


src/SB/Core/x/xCollide.cpp
```diff
@@ -251,7 +264,398 @@ void xParabolaRecenter(xParabola* p, F32 newZeroT)
...
+}
+
+// Get these assignment operators to exist
+static void tempReferenceWeakDefinitions(RwLine* a, nodeInfo* b, RwBBox* c, xCollis* d)
+{
+    *a = *a;
+    *b = *b;
+    *c = *c;
+    *d = *d;
+}
```
JoshSanch: This particular change I'm not sure about. Personally, I'm not a fan of these short-term matches just for getting the matchingness percentage up since it creates cleanup work for later on. I'm open to discussing it though if someone feels differently.

JoshSanch (commented):
Looks good overall! Just one part that I think needs some discussion.

src/SB/Core/x/xCollide.cpp
JoshSanch: @stravant For the sake of getting this in, please remove this function in line with the recently merged guidelines (#554) on avoiding placeholder functions like this for short-term matchingness increases.

JoshSanch (changes requested):
Please see [prior comment](https://github.com/bfbbdecomp/bfbb/pull/567#discussion_r2103647881)

src/SB/Core/x/xCollide.cpp
stravant: Yeah, I don't need this anymore, I was just trying out whether that worked as an approach to make it link.

JoshSanch:
Can you rebase your branch onto latest main? I'm hoping it will resolve this issue for me in case there's some conflicts

image

JoshSanch (approved):
Lgtm! Thanks :)

PR #568, opened 2025-05-22
zNPCTypeBossPlankton, zNPCTypeBossSB2, zNPCTypeDutchman: register_tweaks work

energydrink02:
Add remaining register_tweaks functions in npc boss files


JoshSanch (approved):
Looks great! Per Discord discussion, I share your concern around having to redefine auto_tweak in a bunch of places, but that's out of scope for this PR. Once this goes in, I'll explore some possibilities to reduce the redundancies there.

PR #569, opened 2025-05-24
xFX progress

stravant:
* I'm pretty confident that `RwEngineInstance` has type `RwGlobals`. There's only one match for it in these changes, but that match is pretty convincing, and I checked cross-checked a couple of other usages which also have the argument counts of the function pointers called in it line up correctly.


src/SB/Core/x/xFX.cpp
```diff
@@ -367,6 +365,47 @@ static RpAtomic* AtomicSetShininess(RpAtomic* atomic, void* data)
...
 }
 
+void AtomicSetEnvMap(RpAtomic*, void*)
+{
+
+}
+
+RpAtomic* xFXAtomicEnvMapSetup(RpAtomic* atomic, U32 aid, F32 shininess)
+{
+    if (void* asset = xSTFindAsset(aid, NULL))
```
JoshSanch: This is another one for the style guidelines - please move this assignment outside of the conditional. Ghidra will create really difficult-to-read assignments for conditons more complex than this, so we keep assignments outside of conditons overall to enhance their readability across the project.

JoshSanch (changes requested):

src/SB/Core/x/xFX.cpp
stravant: Ah, that makes sense as a reason to disallow, thanks for the explanation.

PR #570, opened 2025-05-25
SDK Merge

LivewireCB:
All Dolphin SDK files are added and building. With the majority of them being close to Matching.
- 5/24/25
- 84.38% Matched
- 24.90% linked


config/GQPE78/splits.txt
JoshSanch, line 1: Can you share the reasoning behind the need to rearrange the filepaths for the SDK source files?

config/GQPE78/symbols.txt
```diff
@@ -11990,7 +11990,7 @@ axDspSlave = .data:0x802B2000; // type:object size:0x1EC0 scope:global
 ...data.0 = .data:0x802B3EC0; // type:label scope:local
 @1 = .data:0x802B3EC0; // type:object size:0x46 scope:local data:string
 ResetFunctionInfo = .data:0x802B3F08; // type:object size:0x10 scope:local
-CardData = .data:0x802B3F20; // type:object size:0x160 scope:local
+CardData = .data:0x802B3F20; // type:object size:0x160 scope:local align:32
```
JoshSanch, line 11993: Why was this new alignment introduced?

configure.py
```diff
@@ -159,7 +159,7 @@
     "-I include",
     f"-I build/{config.version}/include",
     f"--defsym BUILD_VERSION={version_num}",
-    f"--defsym VERSION_{config.version}",
+    f"--defsym VERSION_={config.version}",
```
JoshSanch, line 162: Was there a particular bug that you encountered that motivated this change?

configure.py
```diff
@@ -200,6 +200,9 @@
     "-str reuse",
     "-multibyte",  # For Wii compilers, replace with `-enc SJIS`
     "-i include",
+    "-i libs/PowerPC_EABI_Support/include",
+    "-i include/dolphin",
+    "-i libs",
     f"-i build/{config.version}/include",
     f"-DBUILD_VERSION={version_num}",
     f"-DVERSION_{config.version}",
```
JoshSanch, line 208: This might need the same fix that you included above.
```suggestion
    f"-DVERSION_={config.version}",
```

JoshSanch (commented):
Absolutely fantastic work. This wasn't strictly necessary for the project to be considered a success in the long-term but having a comprehensive decompilation of the game and all of its dependencies is a worthy goal.

Some discussion questions for follow-up for documentation purposes - I mentioned in the Discord but there are some project structure issues that I've determined as out-of-scope for this PR related to our `include` and the introduced `libs` folders, but that work does not need to block this PR.

config/GQPE78/splits.txt
LivewireCB: To create one cohesive environment for the .c files to live in. They don't belong in the same dir as our game code.
since we have 4 library's it makes sense to put them in one lib folder then their respective subfolder.

config/GQPE78/symbols.txt
LivewireCB: CardBios was giving a checksum error even though the file was 100% matched. The data section was misaligned and off by 8 bytes so I realigned it to fix the error.

configure.py
LivewireCB: Fixed a --defsym compiler error. It was never brought to light because our main project doesn't use it (I believe).

configure.py
LivewireCB: Nah this is fine, its just the include paths.

JoshSanch:
Note on regressions: Implemented code in ngcrad3d was straight up incorrect and did not match SDK definitions, so regressions are expected but also just the realization of those issues. For that reason, I am merging anyway.

Issue #571, opened 2025-05-25
Migrate Game Dependencies to `Extern` Top-Level Directory


JoshSanch:
At the project structure level - we shouldn't really have an include directory at all at the moment since the BfBB decomp isn't a library for other projects to import. The [Melee decomp project](https://github.com/doldecomp/melee/tree/master) follows the C++ project standard of having external dependencies contained in a extern/external folder and then each dependency is a subdirectory under that, for example.

We should refactor the project such that all external dependencies (Dolphin SDK, Renderware SDK, Bink, PowerPC EABI, etc.) are contained within
a top-level directory `extern`. This will require making lots of small changes to compiler paths and such.

Some relevant links
https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1204r0.html
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines

PR #582, opened 2025-05-31
Project Restructuring

LivewireCB:
Restructured the filesystem to be more in line with C++ standards.


JoshSanch:
Closes #571 . We've decided to prioritize sticking with folder conventions from other decomp projects while the project is in flight - understanding that as the project gets to closer to completion, we should migrate the structure to be more friendly to modders and project forks by moving library source and includes (Renderware SDK, Dolphin SDK, BINK, etc.) to an `extern` top-level directory.

PR #588, opened 2025-08-10
zAssetTypes done for now

SquareMan:
Looks to be Equivalent but I haven't tested it in Dolphin. Holding off on marking it Equivalent in configure.py for now.


JoshSanch (approved):

SquareMan:
After playing around in Jellyfish Fields the game seems normal so I'll mark this one as equivalent

PR #589, opened 2025-08-12
zNPCTypeBossPlankton & zNPCTypeBossSB2 work

LivewireCB:
Did quite a few functions in zNPCTypeBossPlankton  and zNPCTypeBossSB2

### DIFF
--- a/src/SB/Game/zNPCTypeBossPlankton.cpp
+++ b/src/SB/Game/zNPCTypeBossPlankton.cpp
@@ -59,54 +59,115 @@ namespace
         U32 flags;
     };
 
-    static U32 sound_asset_ids[6][10];
-    static sound_data_type sound_data[6];
-    static const sound_asset sound_assets[29] =
+    struct bolt;
+
+    struct effect_data
     {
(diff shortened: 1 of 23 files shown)


src/SB/Game/zNPCTypeBossPlankton.cpp
```diff
@@ -988,6 +1072,33 @@ void zNPCBPlankton::Init(xEntAsset* asset) //66%
     zNPCBPlankton::aim_gun(play, &gun_tilt, &move.dest, 0);
 }
 
+void zNPCBPlankton::Setup()
+{
+    U32 tmpVar;
+
+    zNPCBoss::Setup();
+    zNPCBPlankton::setup_beam();
+    tmpVar = xStrHash("NPC_NEWSCASTER");
```
SquareMan, line 1081: It looks like the regression is because this string comes before all the debug tweak strings the rodata and now the diff algorithm doesn't see the tweak string data as matching. I'm not gonna worry about it for now since this file is clearly very WIP but it might be as simple as a function ordering issue.

SquareMan (approved):

PR #593, opened 2025-09-04
zGrid 93%

bluisblu:


src/SB/Core/x/xGrid.h
```diff
@@ -62,9 +62,12 @@ typedef S32 (*GridEntCallback)(xEnt*, void*);
 extern volatile S32 gGridIterActive;
 
 void xGridBoundInit(xGridBound* gridb, void* data);
+void xGridInit(xGrid* grid /* r18 */, const xBox* bounds /* r2 */, U16 nx /* r17 */,
+               U16 nz /* r16 */, U8 ingrid_id /* r2 */);
```
SquareMan: Please cleanup these register comments from the DWARF. They are interpreted from PS2 data and thus make no sense for us.

src/SB/Game/zGrid.h
```diff
@@ -5,6 +5,9 @@
 
 #include "xEnt.h"
 #include "xGrid.h"
+#include "xMath.h"
+
+#include <PowerPC_EABI_Support\MSL_C\MSL_Common\cmath>
```
SquareMan: Let's remove the includes from the header as well since there's nothing in the header that needs them. If necessary move them to the cpp file that actually uses them (likely zGrid.cpp) but if they aren't actually used they should just be removed outright.

SquareMan: Additionally, I might be wrong so ignore this if it doesn't work but I"m pretty sure the cmath include should just be this:
```suggestion
#include "xMath.h"

#include <cmath>
```

SquareMan (changes requested):
Functions are looking good but there's a few minor things that need addressing before merge.

src/SB/Game/zGrid.h
bluisblu: Nope, full path is necessary unfortunately.

bluisblu: Removed, wasn't needed.

src/SB/Core/x/xGrid.h
bluisblu: Done 👍

src/SB/Game/zGrid.h
SquareMan: Ok, this may be a project configuration issue I should look into.

SquareMan: It should still be  moved to zGrid.cpp though

PR #594, opened 2025-09-08
ztalkbox::stop_wait 100%

LSLees:


src/SB/Game/zTalkBox.cpp
```diff
@@ -312,14 +313,22 @@ void ztalkbox::stop_talk()
...
+    
+    if (shared.active != this)
     {
-        //shared.unk8D48 |= unk;
+        return;
     }
+    
+    shared.wait_event_mask = shared.wait_event_mask | x;
+    return;
+
```
tgsm:
```suggestion
    if (shared.active == this)
    {
        shared.wait_event_mask |= x;
    }
```

SquareMan:
In addition to what tgsm has reviewed I would suggest setting up clang-format for your editor and having it run on file save. You have a lot of extra empty lines that would be removed by a formatting pass.

PR #596, opened 2025-09-09
zNPCMessenger work and Misc fixes

LivewireCB:
zNPCMessenger work and Misc fixes


src/SB/Core/gc/iModel.cpp
```diff
@@ -23,45 +22,55 @@ void* GetHierarchy(RpAtomic* frame)
...
 void iModelUnload(RpAtomic* userdata)
 {
-    RpClump* clump;
-    RwFrame* frame = 0;
-    RwFrame* root;
+    // Not really a hard function.
+    // Don't know why frame and root exist
+    RpClump* clump; // r17
+    RwFrame* frame; // r16
+    RwFrame* root; // r2
```
SquareMan: Let's remove these register comments. I'm assuming these are taken from the dwarf as they don't seem match the metrowerks pattern and the unused variables shouldn't be given a register in the first place.

config/GQPE78/config.yml
```diff
@@ -5,4 +5,5 @@ symbols: config/GQPE78/symbols.txt
 splits: config/GQPE78/splits.txt
 symbols_known: true
 fill_gaps: false
-ldscript_template: config/GQPE78/ldscript.tpl
\ No newline at end of file
+ldscript_template: config/GQPE78/ldscript.tpl
+write_asm: true
```
SquareMan: What does this do? We probably don't want it in this PR

src/SB/Core/x/xFX.cpp
```diff
@@ -365,9 +369,14 @@ static RpAtomic* AtomicSetShininess(RpAtomic* atomic, void* data)
...
 }
 
-void AtomicSetEnvMap(RpAtomic*, void*)
+static RpAtomic* AtomicSetEnvMap(RpAtomic* atomic, void* data)
 {
-
+    RpMatFXAtomicEnableEffects(atomic);
+    if ((atomic->geometry) != 0)
+    {
+        RpGeometryForAllMaterials((atomic->geometry), MaterialSetEnvMap2, data);
```
SquareMan: nit: remove the redundant parentheses around `(atomic->geometry)`

src/SB/Core/x/xFX.cpp
```diff
@@ -663,11 +674,67 @@ void xFXShineRender()
...
+}
+
 void xFXAuraRender()
 {
+    // TODO: Fix function
+    // Honestly the closest I can get this function.
+    // fogstate is in the dwarf but currently isn't used in the function.
+
+    S32 fogstate; // r29+0x3C
+    _xFXAura* ap; // r17
```
SquareMan: Remove these register comments as well.

src/SB/Core/x/xFX.cpp
```diff
@@ -909,22 +995,105 @@ void xFXanimUV2PSetTranslation(const xVec3*)
...
+    RwRenderStateSet(rwRENDERSTATECULLMODE, (void*)0x8);
+
+    return atomic;
 }
 
 void xFXRibbonRender()
 {
+    xFXRibbon* prev; // r19
+    U32 i; // r18
+    xFXRibbon* ribbon; // r17
```
SquareMan: more register comments

src/SB/Core/x/xParSys.cpp
```diff
@@ -1,3 +1,50 @@
...
+        iParMgrRenderParSys_Sprite(data, ps);
+    }
+}
+
+void xParCmdTexInit(xParCmdTex* tex)
+{
+    // TODO: Function either needs casts, or its just float meme
+    tex->unit_count = tex->rows * tex->cols;
+    tex->unit_width = (tex->x2 - tex->x1) / (tex->cols - 4503599627370496.0);
+    tex->unit_height = (tex->y2 - tex->y1) / (tex->rows - 4503599627370496.0);
```
SquareMan: this float value is used as a compiler implementation detail when converting ints to float (or vice-versa I don't 100% remember). I think the code is just `(tex->x2 - tex->x1) / tex->cols` here.

src/SB/Core/x/xParSys.cpp
```diff
@@ -1,3 +1,50 @@
...
+    xParSysInit((xBase*)b, (xParSysAsset*)tasset);
+}
+
+void xParSysInit(xBase* b, xParSysAsset* tasset)
+{
+}
+
+void xParSysSetup(xParSys* t)
+{
+    if ((t != 0) && (t->link != 0) && (t->link->param[1]))
```
SquareMan: More redundant parentheses here I'd like to see removed.

src/SB/Core/x/xParSys.h
```diff
@@ -35,10 +37,21 @@ struct xParSys : xBase
...
+void xParCmdTexInit(xParCmdTex* tex);
 void xParSysInit(void* b, void* tasset);
+void xParSysInit(xBase* b, xParSysAsset* tasset);
 void xParSysSetup(xParSys* t);
 void xParSysExit(xParSys* t);
 void xParSysRender(xBase* b);
+S32 xParSysEventCB(xBase*, xBase*, U32, F32*, xBase*);
 void xParSysUpdate(xBase* to, xScene*, F32 dt);
 
+WEAK U8 using_ptank_render(const xParSysAsset&);
```
SquareMan: We don't really use the WEAK macro anymore. This function should probably be marked either `inline` or `static`.

src/SB/Game/zEntPlayer.cpp
```diff
@@ -79,6 +79,7 @@ static S32 sPlayerAttackInAir;
 #define MAX_DELAYED_SOUNDS 8
 static zDelayedStreamSound sDelayedSound[MAX_DELAYED_SOUNDS];
 static zPlayerSndTimer sPlayerStreamSndTimer[ePlayerStreamSnd_Total] = {};
+//static PlayerStreakInfo sStreakInfo[3][4] = {};
```
SquareMan: Why is this commented out?

src/SB/Game/zEntPlayer.cpp
```diff
@@ -3698,6 +3699,21 @@ void zEntPlayer_Load(xEnt* ent, xSerial* serial)
     return;
 }
 
+static void zEntPlayer_StreakFX(xEnt* ent, F32)
+{
+    S32 i; // r19
+    S32 p; // r5
+    S32 cp; // r16
```
SquareMan: ditto register comments

src/SB/Game/zMain.cpp
```diff
@@ -740,6 +739,13 @@ void zMainReadINI()
     void* buf;
     U32 size;
     xIniFile* ini;
+    uint local_18[2];
```
SquareMan: This looks like it should just be a U32. It's probably the length of the resulting string but I can't check `iFileLoad` right now.

src/SB/Game/zNPCMessenger.cpp
```diff
@@ -23,12 +25,23 @@ void zNPCMsg_SceneFinish()
     return zNPCMsg_SceneReset();
 }
 
-void zNPCMsg_SendMsg(en_NPC_MSG_ID, U8)
+void zNPCMsg_SendMsg(en_NPC_MSG_ID msgevent /* r17 */, U32 unk0)
```
SquareMan: Register Comment

src/SB/Game/zNPCTypeDutchman.cpp
```diff
@@ -164,6 +161,72 @@ namespace
...
+    static zParEmitter* snot_emitter;
+    static xParEmitterCustomSettings snot_emitter_settings;
+    static zParEmitter* slime_emitter;
+    static xParEmitterCustomSettings slime_emitter_settings;
+    static zParEmitter* hand_trail_emitter;
+    static zParEmitter* blob_emitter;
+
+    static void init_sound()
+    {
+        U32 total; // r11
```
SquareMan: register comment.

src/SB/Game/zNPCTypeDutchman.cpp
```diff
@@ -976,22 +1039,242 @@ xAnimTable* ZNPC_AnimTable_Dutchman()
...
+void zNPCDutchman::Init(xEntAsset* asset)
+{
+    // Function is at 60%
+    // Laser_texture and m both need to be used?
+
+    char* scorch_name[2]; // r29+0x58
+    S32 i; // r18
+    RwTexture* laser_texture; // r2
+    S32 model_index; // r3
+    xModelInstance* m = model; // r4
```
SquareMan: Register comments

src/SB/Game/zNPCTypeDutchman.cpp
```diff
@@ -976,22 +1039,242 @@ xAnimTable* ZNPC_AnimTable_Dutchman()
...
 
+void zNPCDutchman::RenderExtra()
+{
+    S32 oldzwrite; // r29+0x6C
+    S32 oldztest; // r29+0x68
+    U32 oldsrcblend; // r29+0x64
+    U32 olddestblend; // r29+0x60
+    U8 oldcmp; // r17
+    xModelInstance* m; // r16
+    U8 haloing; // r17
```
SquareMan: Register comments

src/SB/Game/zNPCTypeKingJelly.cpp
```diff
@@ -1314,6 +1319,38 @@ void zNPCKingJelly::BUpdate(xVec3* pos)
     zNPCCommon::BUpdate(pos);
 }
 
+S32 zNPCKingJelly::SysEvent(xBase* from, xBase* to, U32 toEvent, const F32* toParam,
+                            xBase* toParamWidget, S32* handled)
+{
+    U32 uVar1;
```
SquareMan: A better name can be inferred.
Example:
```suggestion
    U32 ret;
```

src/SB/Game/zNPCTypeKingJelly.cpp
```diff
@@ -1314,6 +1319,38 @@ void zNPCKingJelly::BUpdate(xVec3* pos)
...
+        uVar1 = 1;
+    }
+    else
+    {
+        if (toEvent < 0x1b9)
+        {
+            if (toEvent == 0x1b5)
+            {
+                start_fight();
+                goto LAB_8014808c;
```
SquareMan: I'm pretty sure these goto's are just `return 1;`. If this function doesn't match yet or this suggestion otherwise doesn't work for some reason we can leave it as is.

SquareMan (changes requested):
Looks pretty good, some cleanup required. A note about all the "register comments": if any of these comments were derived from the actual assembly rather than just being copied from dwarf I'll leave it to your discretion on if they should be left in. Especially if the function they're in aren't matching yet.

config/GQPE78/config.yml
LivewireCB: I mistakenly left this in on accident. Was just testing some stuff. Removed.

src/SB/Core/x/xParSys.cpp
LivewireCB: This was completely correct, and actually matched the function.

src/SB/Core/x/xParSys.h
LivewireCB: Was only using this to signify that it should be weak in the cpp file. Will just remove.

src/SB/Game/zEntPlayer.cpp
LivewireCB: I added it, then forgot to remove when I didn't use it. Will just uncomment instead of removing.

src/SB/Core/x/xParSys.h
SquareMan: I understand. We can leave it alone for now but the point was that `inline` is what cause the symbol to be weak. I think sometimes `static` functions become weak as well, but I don't quite perfectly remember the exact situation I'm thinking of.

SquareMan (approved):
Looks good now.

PR #597, opened 2025-09-23
Work on zNPCType files

bluisblu:
Mostly updating stuff to use zNPCSndLists directly, however I did also get carried away with overhauling zNPCTypeTiki


JoshSanch:
Huge work

Thanks!

PR #598, opened 2025-09-23
zNPCDutchman work

LivewireCB:
Pretty much done with the file.

Creating new branch after this pr.


JoshSanch (changes requested):
Can you investigate the register_tweaks regression? That seems like a real regression and it may be based on the struct changes.

PR #599, opened 2025-09-26
set_alpha_blend 100%, abandon_slugs 100%, choose_hand 100%, NewTime 37%

LSLees:
newtime is in progress, can remove if needed


JoshSanch (changes requested):
In general, we try to avoid merging heavily WIP code, since it can cause confusion for other contributors. Please take out the NewTime contribution - everything else is looking good though!

JoshSanch (approved):
LGTM!

PR #619, opened 2025-10-22
iModel work

escape209:
Some are not 100% but progress is progress.

### DIFF
--- a/src/SB/Core/gc/iModel.cpp
+++ b/src/SB/Core/gc/iModel.cpp
@@ -1,18 +1,36 @@
 #include "iModel.h"
 
 #include <types.h>
+#include <rpskin.h>
+#include <rpmatfx.h>
+
 #include "zAssetTypes.h"
+#include "xMathInlines.h"
+
+extern RwGlobals* RwEngineInstance;
+static RpAtomic* sLastMaterial;
+static U32 sMaterialFlags;
+static U32 sMaterialIdx;
+static RwTexture* sMaterialTexture[16];
+static U32 gLastAtomicCount;
+static RpAtomic* gLastAtomicList[256];
+
+static RpLight* sEmptyDirectionalLight[4];
+static RpLight* sEmptyAmbientLight;
+
+int iModelHack_DisablePrelight;
 
 RwFrame* GetChildFrameHierarchy(RwFrame* frame, void* data)
 {
     RpHAnimHierarchy* hierarchy = RpHAnimFrameGetHierarchy(frame);
     if (hierarchy == 0)
     {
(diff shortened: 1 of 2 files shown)


src/SB/Core/gc/iModel.cpp
```diff
@@ -26,51 +47,314 @@ void* GetHierarchy(RpAtomic* frame)
...
+    RpGeometry* geom;
+    RwMatrixTag *pAnimOldMatrix;
+    RwFrame* frame;
+
+    iVar1 = (RpHAnimHierarchy *)GetHierarchy(model);
+    if ((S8)init == false)
+    {
+        draw_all = 1;
+        init = true;
+    }
```
tgsm:
```suggestion
    static int draw_all = 1;
```

escape209: Assuming the variable `init` should be removed, too?

JoshSanch: Yep, `init` is an implicit variable generated by the compiler to handle the static initialization

include/rwsdk/rwplcore.h
```diff
@@ -673,6 +673,29 @@ struct RwMatrixTag
     // Offset: 0x30
     RwV3d pos;
     RwUInt32 pad3;
+
+	RwMatrixTag& operator=(const RwMatrixTag& rhs)
```
SquareMan: I'm not so sure about this. The S32 casts look like they're necessary because the asm is likely a compiler generated memcpy. This happens when the compiler generates the assignment function implicitly, meaning it is not very declared or defined.

Additionally render ware is a C library AFAIK so it wouldn't have supported this feature.

What happens when you remove this entirely?

src/SB/Core/gc/iModel.cpp
```diff
@@ -1,18 +1,36 @@
...
     if (hierarchy == 0)
     {
-        RwFrameForAllChildren(frame, GetChildFrameHierarchy, data);
+        RwFrameForAllChildren(frame, &GetChildFrameHierarchy, data);
+		return frame;
     }
     else
     {
-        data = hierarchy;
+        *(int *)data = (int)hierarchy;
```
SquareMan: Use S32 instead of int for bfbb code

SquareMan (changes requested):

include/rwsdk/rwplcore.h
escape209: Oh it's fine 🤔
Yeah this should just not have been added. I think I added it out of habit since I've added something like this many times. In fact I'll probably go over any other operator defs afterwards.

PR #621, opened 2025-10-24
zEntCruiseBubble: fix some signatures + couple matches

escape209:


src/SB/Game/zEntCruiseBubble.cpp
```diff
@@ -859,6 +898,45 @@ namespace cruise_bubble
             }
         }
 
+		void init_states()
+		{
+			static state_player_halt player_halt;
+			shared.states[0] = (state_type*)&player_halt;
```
SquareMan: Are these casts on all the states necessary? `state_type` is the base class of all of these right?

src/SB/Game/zEntCruiseBubble.cpp
```diff
@@ -961,6 +1039,117 @@ namespace cruise_bubble
...
+            // scheduling off
+            F32 vel_frac = ((state_missle_fly*)shared.states[STATE_MISSLE_FLY])->vel /
+                           current_tweak->missle.fly.max_vel;
+
+            hud.uv_wind.offset_vel.assign(current_tweak->hud.wind.du, current_tweak->hud.wind.dv);
+            hud.uv_wind.offset_vel *= vel_frac;
+            hud.model.wind->Alpha = vel_frac;
+            hud.uv_wind.update(dt);
+
+            // sheduling off for i and zEntCruiseBubble_f_n1_0
```
SquareMan, line 1063: You could try replacing the externed floats in this file with the proper float literals, it may help the scheduling. Not gonna block the merge on this though since there's a lot and they are still leftover from the older build setup.

Edit: looks like you only moved this function up. Just ignore this if you don't wanna mess with it for now.

SquareMan (changes requested):

src/SB/Game/zEntCruiseBubble.cpp
escape209: No actually. Must've had some explicit conversion error pop up at some point and got confused.

src/SB/Game/zEntCruiseBubble.cpp
escape209: Was actually about to do this too, but thought that probably deserved a PR on its own.

JoshSanch (approved):

PR #622, opened 2025-10-24
Most of xSpline

daft7:
Pretty good matches for xSpline, definitely some tough ones in there.

### DIFF
--- a/src/SB/Core/x/xSpline.h
+++ b/src/SB/Core/x/xSpline.h
@@ -8,6 +8,10 @@ struct xCoef
     F32 a[4];
 };
 
+static F32 sBasisUniformBspline[4][4];
+static F32 sBasisBezier[4][4];
+static F32 sBasisHermite[4][4];
+
 struct xCoef3
 {
     xCoef x;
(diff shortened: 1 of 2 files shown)


src/SB/Core/x/xSpline.h
```diff
@@ -8,6 +8,10 @@ struct xCoef
     F32 a[4];
 };
 
+static F32 sBasisUniformBspline[4][4];
+static F32 sBasisBezier[4][4];
+static F32 sBasisHermite[4][4];
```
JoshSanch: Did you have a reason for putting these into the header instead of the cpp file?

JoshSanch (changes requested):
Please run both files through clang-format, but otherwise it's looking great. Good work!

src/SB/Core/x/xSpline.h
daft7: In my mind, it's overhead and goes in the h but otherwise no.

JoshSanch: Since these are static, let's go ahead and move them to the cpp. Variables that could be exposed to other files go into the header, but ones that only live in the compilation unit (xSpline.cpp) should go into the cpp. Static inherently would be only linked once to the xSpline compilation unit so let's go with that.

PR #625, opened 2025-10-27
xMovePoint equivalent match

bluisblu:
As discussed in voice chat, `xVec3::cross` had an initialized `xVec3 v` which caused data to leak into other files that included it, so this will cause a regression necessary for matching the xMovePoint file, which will have to be solved some other way.

Dolphin test seems fine, and the only reason it's not 100% is a regswap in the remaining function.

### DIFF
--- a/src/SB/Core/x/xMovePoint.cpp
+++ b/src/SB/Core/x/xMovePoint.cpp
@@ -1,12 +1,13 @@
 #include "xMovePoint.h"
 
+#include "xMathInlines.h"
 #include <types.h>
 
 #include "xMath.h"
 #include "xScene.h"
 #include "xMemMgr.h"
 
-F32 xVec3Hdng(xVec3* a, const xVec3* b, const xVec3* c);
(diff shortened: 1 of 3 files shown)


src/SB/Core/x/xMovePoint.cpp
```diff
@@ -106,33 +107,29 @@ void xMovePointSplineSetup(xMovePoint* m)
...
 
-// If you uncomment the numPoints variable then this function is a perfect match
-// minus ordering. In the original assembly some variable fetches are lifted to
-// places earlier in the assembly listing than what this comiles to for some
-// reason.
-// The r27-30 registers used to hold the main varibles of the function are all
-// perfect matches with this code.
-extern float xMovePoint_float_0;
-F32 xMovePointGetNext(xMovePoint* m, xMovePoint* prev, xMovePoint** next, xVec3* hdng)
+// Scratch: https://decomp.me/scratch/JFBBX
```
JoshSanch: No need to link a scratch in the comments here, let's clean it up before merging.

JoshSanch (changes requested):

PR #626, opened 2025-10-27
Fix `xVec3::cross`

bluisblu:
Restore the logic that made it match as well as it did with initialization without causing a `.rodata` leak.


JoshSanch (approved):
Finally our ancestors (me) can rest in peace

PR #628, opened 2025-10-28
xBehaveMgr work

escape209:
Pls work this time

### DIFF
--- a/src/SB/Core/x/xBehaviour.h
+++ b/src/SB/Core/x/xBehaviour.h
@@ -69,8 +69,8 @@ enum en_psynote
 
 enum en_xpsytime
 {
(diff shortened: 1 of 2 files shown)


src/SB/Core/x/xBehaviour.h
```diff
@@ -67,6 +67,12 @@ enum en_psynote
     PSY_NOTE_FORCE = 0x7fffffff
 };
 
+enum en_xpsytime
```
JoshSanch, line 70: Please adjust this to adhere to project code convention guidelines - avoid explicitly setting enum values if the enum has typical value assignments (starts at zero and increments by 1).

JoshSanch (changes requested):

PR #630, opened 2025-10-30
zEntPlayer work

escape209:


src/SB/Game/zEntPlayer.cpp
```diff
@@ -3905,6 +4014,16 @@ void zEntPlayer_GiveLevelPickupCurrentLevel(S32 quantity)
     }
 }
 
+xVec3 sDriveVel;
```
JoshSanch: This should be static and declared at the top of the file.

src/SB/Game/zEntPlayer.h
```diff
@@ -446,7 +446,7 @@ void zEntPlayerControlOff(zControlOwner owner);
 void zEntPlayerControlOn(zControlOwner owner);
 
 void zEntPlayer_StoreCheckPoint(xVec3* pos, F32 rot, U32 initCamID);
-S32 load_talk_filter(U8* filter, xModelAssetParam* params, U32 params_size, S32 max_size);
+static S32 load_talk_filter(U8* filter, xModelAssetParam* params, U32 params_size, S32 max_size);
```
JoshSanch: Static functions that aren't part of a class should not be declared in the header, which is generally reserved for global variables, functions, and struct definitions. Please remove the declaration here and redeclare at the top of the cpp if needed to compile.

JoshSanch (changes requested):

src/SB/Game/zEntPlayer.cpp
escape209: With static it'd be in the wrong section (should be in .comm), but it should be up top.

src/SB/Game/zEntPlayer.cpp
```diff
@@ -212,6 +212,9 @@ static F32 sBubbleBowlTimer;
 static U32 sSpatulaGrabbed;
 S32 gWaitingToAutoSave;
 
+xMat4x3 sCameraLastMat;
```
JoshSanch, line 215: If these are static, let's add the static keyword. Otherwise looks good

escape209: The DWARF says they aren't, and adding the static keyword sticks them in .bss instead of .comm, so I don't think they are

JoshSanch: Ok, love that Heavy Iron named them with static naming convention anyways :)

Does objdiff list them as global or local variables?

escape209: Extern

src/SB/Game/zEntPlayer.cpp
```diff
@@ -3517,40 +3525,40 @@ S32 zEntPlayer_IsSneaking()
...
+        for (S32 j = 0; j < found; j++)
+		{
+            if ((S32)*non_choice - 1 == i)
             {
                 skip = true;
                 break;
             }
-        }
-
+            non_choice++;
```
SquareMan, line 3547: Question: Did you try using `non_choices[j]` instead of manually iterating the array since you mention that `non_choice` isn't in the DWARF?

src/SB/Game/zEntPlayer.cpp
```diff
@@ -3639,7 +3693,7 @@ void zEntPlayerPreReset()
 F32 ComputeFudge(F32 a, F32 b)
 {
     F32 min = MIN(a, b);
-    a = (min - 0.075f) / -0.175f;
+    a = (min - -0.175f) / 0.074999996f; // Will not match with 0.075f.
```
SquareMan, line 3696: This could be `a = (min - -0.175f) * 13.33333333f`, try to experiment with that and how to round the .33333... and see if it even becomes a division instead of multiplication.

src/SB/Game/zEntPlayer.cpp
```diff
@@ -6630,6 +6790,180 @@ xAnimTable* zEntPlayer_AnimTable()
...
+
+static void dampen_velocity(xVec3& v1, const xVec3& v2, F32 f)
+{
+	F32 f0 = v1.x * v2.x;
+
+	S32 v1_old[3];
+
+	v1_old[0] = *(S32*)(&v1.x);
+	v1_old[1] = *(S32*)(&v1.y);
+	v1_old[2] = *(S32*)(&v1.z);
```
SquareMan: Did you try `S32 v1_old = {v1.x, v1.y, v1.z};`? There must be some automatic copy thing going on for the compiler to use the normal register for the copy instead of float registers.

SquareMan: Also are we sure it's an S32 and not `xVec3 v1_old = v1`? Maybe just leave a //FIXME  comment here in case you can't figure it out for now.

src/SB/Game/zEntPlayer.cpp
escape209: Strangely I didn't, _but_ I noticed that doing that actually flips the operands of the comparison on line 0x88, while the extra variable puts them in the correct order. I feel like your suggestion is correct, though and this is just compiler weirdness again.

src/SB/Game/zEntPlayer.cpp
escape209: No luck, but I realised `(3.0f / 40.0f)` instead of `0.074999996f` also works. Not really sure if that's really "better" clarity-wise, though.

src/SB/Game/zEntPlayer.cpp
escape209: It was actually an xVec3

src/SB/Game/zEntPlayer.cpp
JoshSanch: I like it better in terms of clarity - usually the fraction is more expressive about the programmer's intent.

src/SB/Game/zEntPlayer.cpp
JoshSanch: Let's opt for the Square suggestion and take the less-matching solution for the sake of likely correctness later on.

PR #636, opened 2025-11-11
Match zVar

escape209:
Some weird thing with a bunch of objects that are initialized but never used, but otherwise 100% matching.


escape209:
Called it "zNPCGlyph". Oops.
Will fix tomorrow

JoshSanch:
This doesn't quite match in objdiff in terms of where the bytes live in each segment, but the resulting binary generates the correct SHA when marked as matching and built using ninja. Since the DOL SHA passes verification, I'm comfortable marking this as Matching. Great work!

PR #638, opened 2025-11-13
Match iMix

escape209:
This PR also deletes two seemingly erroneous files, `iMix.cpp` and `iMix.h` that I don't think should be there. Also corrects use of an iMix function to match.


JoshSanch (approved):
Verified that SHA matches after adding to configure.py, and code changes look good :) merging

PR #639, opened 2025-11-14
xParEmitter and xParEmitterType progress

daft7:
These two files are now pretty well established. Some header tweaks for overhead and all functions now defined with some work done.

### DIFF
--- a/src/SB/Core/x/xParEmitter.cpp
+++ b/src/SB/Core/x/xParEmitter.cpp
@@ -1,3 +1,828 @@
 #include "xParEmitter.h"
 
 #include <types.h>
+#include <zScene.h>
+#include <iModel.h>
+#include <xMathInlines.h>
+#include <xMath.h>
+#include <xGroup.h>
+#include <zGlobals.h>
+#include <xDebug.h>
+
+void add_tweaks(xParEmitter& pe)
+{
+}
+
+S32 xParInterpConvertInterpMode(xParInterp* p)
+{
+    U32 interp_val;
+
+    interp_val = p->interp;
+    if (interp_val < 8)
+    {
+        return interp_val;
+    }
+    if (interp_val == xStrHash("s_ConstA_004ff970"))
+    {
+        return 0;
+    }
+    if (interp_val == xStrHash("s_ConstB_004ff977"))
+    {
+        return 1;
+    }
+    if (interp_val == xStrHash("&DAT_004ff97e"))
+    {
+        return 4;
+    }
+    if (interp_val == xStrHash("s_Cosine_004ff983"))
+    {
+        return 5;
+    }
+    if (interp_val == xStrHash("s_Linear_004ff98a"))
+    {
+        return 3;
+    }
+    if (interp_val == xStrHash("&DAT_004ff991"))
+    {
+        return 7;
+    }
+    if (interp_val == xStrHash("s_Random_004ff996"))
+    {
+        return 2;
+    }
+    return interp_val;
+}
+
+void xParEmitterInit(void* b, void* tasset)
+{
+    xParEmitterInit((xBase*)b, (xParEmitterAsset*)tasset);
+}
+
+void xParEmitterInit(xBase* b, xParEmitterAsset* pea)
+{
+    xLinkAsset* prop;
+    xLinkAsset* t;
+    S32 i;
+    F32 vec_length;
+    F32 temp_rate;
+    F32 temp_oocull;
+    U32 temp_psi;
+
+    xParEmitter* bEmitter = (xParEmitter*)b;
+
+    xBaseInit(bEmitter, (xBaseAsset*)pea);
+    bEmitter->eventFunc = (xBaseEventCB)xParEmitterEventCB;
+    bEmitter->tasset = pea;
+    bEmitter->prop = (xParEmitterPropsAsset*)xSTFindAsset(pea->propID, NULL);
+    if (bEmitter->prop == NULL)
+    {
+        bEmitter->prop = 0;
+    }
+    if (bEmitter->linkCount != 0)
+    {
+        bEmitter->link = bEmitter->link;
+    }
+    else
+    {
+        bEmitter->link = NULL;
+    }
+    prop = (xLinkAsset*)bEmitter->prop;
+    bEmitter->emit_flags = pea->emit_flags;
+    temp_psi = bEmitter->prop->parSysID;
+    if (temp_psi != 0)
+    {
+        bEmitter->parSys = (xParSys*)zSceneFindObject(temp_psi);
+    }
+    else
+    {
+        bEmitter->parSys = NULL;
+    }
+    t = prop;
+    bEmitter->attachTo = 0;
+    bEmitter->emit_volume = 0;
+    for (i = 0; i < 0xE; i += 1)
+    {
+        if (t->param[1] == t->param[2])
+        {
+            t->param[3] = 0U;
+        }
+        else
+        {
+            t->param[3] = (F32)xParInterpConvertInterpMode((xParInterp*)(t->param + 1));
+        }
+        t += 0x14;
+    }
+    bEmitter->rate_mode = (U8)prop->param[3];
+    if (((U8)bEmitter->eventFunc == 2) && (prop->param[1] == prop->param[2]))
+    {
+        bEmitter->rate_mode = 0;
+    }
+    if (bEmitter->rate_mode == 2)
+    {
+        (bEmitter->prop->life).order();
+    }
+    if (bEmitter->rate_mode == 2)
+    {
+        temp_rate = ((prop->param[1] - prop->param[2]) * xurand()) + prop->param[2];
+    }
+    else
+    {
+        temp_rate = prop->param[1];
+    }
+    bEmitter->rate = temp_rate;
+    bEmitter->rate_time = 0.0f;
+    bEmitter->rate_fraction = 0.0f;
+    bEmitter->rate_fraction_cull = 0.0f;
+    if (pea->emit_flags & 2)
+    {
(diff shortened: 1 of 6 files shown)


src/SB/Core/x/xParEmitter.h
```diff
@@ -72,43 +74,54 @@ struct xParEmitterAsset : xBaseAsset
...
 
+void add_tweaks(xParEmitter& pe);
+S32 xParInterpConvertInterpMode(xParInterp* p);
 void xParEmitterInit(void* b, void* tasset);
+void xParEmitterInit(xBase* b, xParEmitterAsset* pea);
 void xParEmitterSetup(xParEmitter* t);
-void xParEmitterDestroy();
-void xParEmitterUpdate(xBase* to, xScene*, F32 dt);
+void xParEmitterReset(xParEmitter* t);
+S32 xParEmitterEventCB(xBase* to1, xBase* to2, U32 toEvent, F32* unused1, xBase* unused2);
```
JoshSanch: There's standard symbol names for these callbacks - please update this to use them. Here's a reference from xGroup:

```cpp
S32 xGroupEventCB(xBase* to, xBase* from, U32 toEvent, const F32* toParam,
                    xBase* toParamWidget);
```

src/SB/Core/x/xParEmitter.h
```diff
@@ -16,27 +16,29 @@ struct xParInterp
 
     void set(F32, F32, F32, U32);
     void set(F32); // Used in zNPCDutchman
+    void order();
+    void operator=(const xParInterp&);
```
JoshSanch: Please add the parameter name you used in the cpp here for parity.

src/SB/Core/x/xParEmitter.cpp
```diff
@@ -1,3 +1,828 @@
...
+    if ((parTo->parSys != NULL) && (parTo->emit_flags & 1))
+    {
+        xParEmitterEmit(parTo, dt);
+    }
+}
+
+void xParEmitterDestroy()
+{
+    xDebugRemoveTweak("Particle Emitters");
+    return;
```
JoshSanch: No need for the explicit return here - please remove for ease of maintenance

src/SB/Core/x/xParEmitter.cpp
```diff
@@ -1,3 +1,828 @@
...
+
+void add_tweaks(xParEmitter& pe)
+{
+}
+
+S32 xParInterpConvertInterpMode(xParInterp* p)
+{
+    U32 interp_val;
+
+    interp_val = p->interp;
```
JoshSanch: Let's do the declaration and assignment on the same line since the variable is initialized immediately.

src/SB/Core/x/xParEmitter.cpp
```diff
@@ -1,3 +1,828 @@
...
+    }
+    if (interp_val == xStrHash("s_Cosine_004ff983"))
+    {
+        return 5;
+    }
+    if (interp_val == xStrHash("s_Linear_004ff98a"))
+    {
+        return 3;
+    }
+    if (interp_val == xStrHash("&DAT_004ff991"))
```
JoshSanch: Is this what objdiff shows as the string value? This seems more like a Ghidra artifact.

src/SB/Core/x/xParEmitter.cpp
```diff
@@ -1,3 +1,828 @@
...
+        bEmitter->parSys = (xParSys*)zSceneFindObject(temp_psi);
+    }
+    else
+    {
+        bEmitter->parSys = NULL;
+    }
+    t = prop;
+    bEmitter->attachTo = 0;
+    bEmitter->emit_volume = 0;
+    for (i = 0; i < 0xE; i += 1)
```
JoshSanch: Please use the integer literal here as opposed to the hex format for clarity. Also, `i++` is the conventional way of expressing `i += 1, return the updated value of i`.

src/SB/Core/x/xParEmitter.cpp
```diff
@@ -1,3 +1,828 @@
...
+        t->group = *(xParGroup**)(t->parSys + 0x20);
+    }
+    if ((t->tasset->attachToID != 0) && (t->tasset->emit_type == '\t'))
+    {
+        temp_base = zSceneFindObject(t->tasset->attachToID);
+        iModelTagSetup(&t->tag, *(RpAtomic**)(*(S32*)&temp_base[2].baseType + 0x10),
+                       *(F32*)&t->tasset->e_entbound, (t->tasset->e_entbound).expand,
+                       (t->tasset->e_entbound).deflection);
+    }
+    temp_base = (xBase*)0x0;
```
JoshSanch: You can just use `NULL` here

src/SB/Core/x/xParEmitter.cpp
```diff
@@ -1,3 +1,828 @@
...
+    xBase* temp_base;
+
+    if (t->parSys != (xParSys*)0x0)
+    {
+        t->group = *(xParGroup**)(t->parSys + 0x20);
+    }
+    if ((t->tasset->attachToID != 0) && (t->tasset->emit_type == '\t'))
+    {
+        temp_base = zSceneFindObject(t->tasset->attachToID);
+        iModelTagSetup(&t->tag, *(RpAtomic**)(*(S32*)&temp_base[2].baseType + 0x10),
```
JoshSanch: Please work with the Discord after making other changes to figure out what this offset should be without using pointer math.

src/SB/Core/x/xParEmitter.cpp
```diff
@@ -1,3 +1,828 @@
...
+                       (t->tasset->e_entbound).deflection);
+    }
+    temp_base = (xBase*)0x0;
+    if (t->tasset->attachToID != 0)
+    {
+        temp_base = zSceneFindObject(t->tasset->attachToID);
+    }
+    t->attachTo = temp_base;
+    if (((t->tasset->emit_type != '\x10') && (t->tasset->emit_type != '\x0F')) &&
+        (t->tasset->emit_type == '\x06'))
```
JoshSanch: Are there existing macros for these emit types that we can use instead of these char literals? If you were able to see what each type was used for, you could create some:

```cpp
#define PAR_EMIT_TYPE_ONE 1
#define PAR_EMIT_TYPE_TWO 2
...
```

src/SB/Core/x/xParEmitter.cpp
```diff
@@ -1,3 +1,828 @@
...
+    t->emit_flags = t->tasset->emit_flags;
+}
+
+S32 xParEmitterEventCB(xBase* to1, xBase* to2, U32 toEvent, F32* unused1, xBase* unused2)
+{
+    xParEmitterCustomSettings sp8;
+
+    switch ((S32)toEvent)
+    {
+    case 0xA:
```
JoshSanch: Please replace ALL of these with the eEvent enums in xEvent if they match. Those are used throughout the codebase.

src/SB/Core/x/xParEmitter.cpp
```diff
@@ -1,3 +1,828 @@
...
+    }
+    if ((custom_flags & 0x100) != 0)
+    {
+        p_tasset->pos = info->pos;
+    }
+    if ((custom_flags & 0x10) != 0)
+    {
+        p->prop->value[0] = info->value[0];
+    }
+    if ((custom_flags & 2) != 0)
```
JoshSanch: Please make the integer flag comparisons consistent with the other hex values by making them hex values to represent the fact that bits are being compared.

src/SB/Core/x/xParEmitter.cpp
```diff
@@ -1,3 +1,828 @@
...
+
+F32 xParInterpCompute(S32 interp_mode, xParInterp* r, F32 time, S32 time_has_elapsed, F32 lastVal)
+
+{
+    F32 val;
+    F32 val_0;
+
+    val_0 = r->val[0];
+
+    val = time;
```
JoshSanch, line 347: Combine the declaration and assignments for these

src/SB/Core/x/xParEmitter.cpp
```diff
@@ -1,3 +1,828 @@
...
+    U32 attachGroupIndex;
+    S32 count;
+    S32 emitAgain;
+    bool marker;
+    S32 group_ptr_chk;
+    U32 temp_maxPar;
+
+    if (pe->parSys == 0)
+    {
+        p = (xPar*)0x0;
```
JoshSanch: Use `NULL` for this one and the one two lines after please

src/SB/Core/x/xParEmitter.cpp
```diff
@@ -1,3 +1,828 @@
...
+        if (rate_has_elapsed > 0)
+        {
+            pe->rate_fraction = pe->rate_fraction - rate_has_elapsed;
+        }
+        if (rate_has_elapsed == 0)
+        {
+            p = (xPar*)0x0;
+        }
+        else
+        {
```
JoshSanch, line 492: Please add a `// TODO` here for cleaning up the logic - works for now but I don't think it needs to be this nested.

JoshSanch (changes requested):
Great, great, great work :)

Left some feedback to clean things up but that's all. Absolutely massive progress.

src/SB/Core/x/xParEmitter.cpp
daft7: I'll mark this as a to do, I should parse the DWARF and see if it's in there and for other useful enums.

daft7: Never mind, wiki had my back https://heavyironmodding.org/wiki/EvilEngine/PARE

JoshSanch (approved):

PR #648, opened 2025-11-17
xHudFontMeter Progress

daft7:
Useful groundwork for xHudFontMeter, all methods for this file.

### DIFF
--- a/src/SB/Core/x/xHudFontMeter.cpp
+++ b/src/SB/Core/x/xHudFontMeter.cpp
@@ -7,7 +7,8 @@
 
 void xhud::font_meter_widget::load(xBase& data, xDynAsset& asset, u32 size_t)
 {
-    widget::init_base((xBase&)*this, *(xBaseAsset*)&data, sizeof(xhud::font_meter_widget) + 0x10);
+    // Stub for now because this fails to build. 100% match locally though.
+    // widget::init_base((xBase&)*this, *(xBaseAsset*)&data, sizeof(xhud::font_meter_widget) + 0x10);
     // new (&this->rc.size)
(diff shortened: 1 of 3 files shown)


src/SB/Core/x/xHudFontMeter.cpp
```diff
@@ -1,3 +1,183 @@
...
+    this->xf.height = this->font.h;
+    this->xf.space = this->font.space;
+
+    this->xf.color = this->font.c;
+    this->xf.clip = screen_bounds;
+}
+
+void xhud::font_meter_widget::destruct()
+{
+    ((xhud::meter_widget*)this)->destruct();
```
SquareMan: Prefer `xhud::meter_widget::destruct()` to call a base class virtual function over manually casting the this pointer.

src/SB/Core/x/xHudFontMeter.cpp
```diff
@@ -1,3 +1,183 @@
...
+
+    (*this).updater(dt);
+    this->xf.id = (this->font).id;
+    this->xf.space = (this->font).space;
+    a = (*this).rc.size.x;
+    this->font.w = a;
+    this->xf.width = a;
+    a = (*this).rc.size.y;
+    this->font.h = a;
+    this->xf.height = a;
```
SquareMan: `(*this).` is the same as `this->`, you should use the arrow operator here for consistency and readability. Additionally the redundant parentheses should be cleaned up (i.e `this->font.id` instead of `(this->font).id`. Same goes for the rest of the additions in this file.

SquareMan (changes requested):

PR #665, opened 2025-11-26
xFont work

bluisblu:


src/SB/Core/x/xFont.cpp
```diff
@@ -2141,18 +2152,12 @@ namespace
...
-    static const xtextbox::callback cb_2293 = { NULL, update_tag_blue, update_tag_blue };
-
-#if 1
-    void parse_tag_blue(xtextbox::jot& a, const xtextbox& tb, const xtextbox&,
-                        const xtextbox::split_tag& ti);
-#else
     void parse_tag_blue(xtextbox::jot& a, const xtextbox& tb, const xtextbox&,
                         const xtextbox::split_tag& ti)
     {
         // todo: uses int-to-float conversion
```
JoshSanch, line 2158: If this matches, can you remove any of the comments saying "uses int-to-float conversion"?

JoshSanch (changes requested):

src/SB/Core/x/xFont.cpp
JoshSanch: resolved in discord - this is just a data thing for now

JoshSanch (approved):

PR #666, opened 2025-11-26
Finish xJSP as Equivalent

energydrink02:
-Finish xJSP as Equivalent (ported Seil's code from his PC port)
-Move RwMemory struct from zAssetTypes.h to rwplcore.h


JoshSanch:
Confirmed through manual assembly analysis that these functions seem to be equivalent and only will not fully match due to compiler scheduling differences. Merging.

PR #667, opened 2025-11-27
zEntCruiseBubble: initial work

escape209:
There's a couple matches here but the main focus is just organization. This PR puts most all the functions in the correct order and namespaces. cb_droplet is definitely not right, even though it matches.


JoshSanch (approved):
I'm thankful for u this Thanksgiving

lgtm

PR #668, opened 2025-11-27
Finish xClumpColl as Equivalent

energydrink02:
Ported code from Seil's PC port. Has some operator= inlines I cannot resolve.

### DIFF
--- a/src/SB/Core/x/xClumpColl.cpp
+++ b/src/SB/Core/x/xClumpColl.cpp
@@ -2,91 +2,889 @@
 
 #include <types.h>
 
-xClumpCollBSPTree* xClumpColl_StaticBufferInit(void* data, U32 param_2)
+U8 xClumpColl_FilterFlags = 0x04;
+
+union IntersectionCallBack
 {
-    U32* header; // unused
+    RpIntersectionCallBackWorldSector sectorCB;
+    RpIntersectionCallBackWorldTriangle worldCB;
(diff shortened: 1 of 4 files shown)


src/SB/Core/x/xClumpColl.cpp
```diff
@@ -2,91 +2,889 @@
...
+};
 
-    U32 numBranchNodes = *(U32*)((int)data + 4);
-    U32 numTriangles = *(U32*)((int)data + 8);
+struct TestSphere
+{
+    RwSphere* sphere;
+    F32 recipRadius;
+};
+
```
JoshSanch, line 44: Let's move these structs into the header file in alignment with the rest of our project conventions.

JoshSanch (changes requested):
Thanks for the effort - just some small clean up to do.

src/SB/Core/x/xClumpColl.cpp
energydrink02: Aren't they local to this specific file only and thus have not to be defined in the header? Otherwise yes I would put them in the header of course.

JoshSanch: I did some follow-up investigation and you seem to be right in that there are redefinitions in other files that are different from the ones here, so they are likely to be locally defined structs. We can go ahead and merge. Good work!

PR #670, opened 2025-11-28
Start of bink sdk

LivewireCB:


tools/project.py
```diff
@@ -1421,7 +1421,7 @@ def generate_objdiff_config(
         "Wii/1.3": "mwcc_43_172",
         "Wii/1.5": "mwcc_43_188",
         "Wii/1.6": "mwcc_43_202",
-        "Wii/1.7": "mwcc_43_213",
+        "ProDG/3.5": "mwcc_43_213",
```
JoshSanch, line 1424: Is this a leftover testing configuration? Not sure why this change is here

JoshSanch (approved):
Super glad to see this work done, and some additionally linked files to boot. Only one comment now that things are working and building ok.

tools/project.py
LivewireCB: ProDG does not have a compiler mapping. So i reused the wii mapping with the prodg compiler, This allows us to create scratches for the ProDG side of the repo. The only caveat being that you have to change the compiler version on each scratch you make to ProDG 3.5

JoshSanch: Can we not add a ProDG mapping rather than removing the correct Wii mapping? Not sure what the technical limitations are here if you could clarify.

LivewireCB: The "mwcc_xx" is the mapping for decomp.me, saying which compiler to use for a scratch. Each mapping you see here are the only ones., and there isn't one for ProgDG. When the compiler maps get updated for the site, we can change this over. This is just a temporary work around since this is out of our hands.

JoshSanch: Alright - can you make a request to Encounter for a ProdDG mapping and tag me on it? Once that's done I will merge this

JoshSanch: FYI @encounter 

https://discord.com/channels/829152115322257436/829152115942228030/1444046495186944061

PR #677, opened 2025-11-29
Completed and Linked xUpdateCull, and xPtankPool

LivewireCB:


src/bink/include/bink.h
```diff
@@ -308,11 +308,13 @@ typedef struct BINKSUMMARY
...
 {
+    // TODO: marked these as volatile to get matches in radcb
+
     u32 FrameNum; // Current frame number
-    u32 FrameRate; // frame rate
+    volatile u32 FrameRate; // frame rate
     u32 FrameRateDiv; // frame rate divisor
     u32 Frames; // frames in this sample period
-    u32 FramesTime; // time is ms for these frames
+    volatile u32 FramesTime; // time is ms for these frames
```
JoshSanch, line 317: Would be good if you could document here or elsewhere why you decided these needed to be marked volatile.

src/SB/Core/x/xPtankPool.cpp
```diff
@@ -243,6 +235,28 @@ namespace
...
+void xPTankPoolSceneEnter()
+{
+    inited = 1;
+    init_groups();
+
+    group_data* it = groups;
+    group_data* end = groups + MAX_PGT;
+    while (it != end)
+    {
+        double f = it->max_size - 0.f;
```
JoshSanch: Do these need to be Heavy Iron-typed? We have F64 for doubles and u32 for unsigned long.

LivewireCB: I honestly don't know as I didn't actually do the function, just fixed it. Will update that now.

src/bink/include/bink.h
LivewireCB: Marked with TODO: in bother bink.h and radcb.c

JoshSanch (approved):

PR #687, opened 2025-11-30
Ported more stuff from:

LivewireCB:
xCamera, xEntMotion, xPad, xSerializer, xString

thank you seil


src/SB/Core/x/xString.cpp
```diff
@@ -47,6 +48,84 @@ U32 xStrHashCat(U32 prefix, const char* str)
...
 }
 
+char* xStrTok(char* string, const char* control, char** nextoken)
+{
+    U8* str;
+    U8* ctrl;
+    U8 map[32];
+    S32 count;
+
+    map[0] = 0;
```
JoshSanch: We should be able to put this in a more concise for loop that unrolls - please try it out.

JoshSanch (changes requested):

PR #692, opened 2025-12-02
xScrFx progress

seilc:
All functions are complete, all data is matching. Mostly scheduling memes.

Also ported RxObjSpace3DVertex/RwIm3DVertex over from Rat, which has it in the DWARF data, so it should hopefully be accurate. It's a bit different from our current one, so I had to edit some places where it's used.


JoshSanch (approved):
Great work! Glad to see some more clarity on RenderWare headers

PR #693, opened 2025-12-02
zNPCMessenger work

bluisblu:


JoshSanch (approved):
Dependency hell has been staved off for another day

PR #704, opened 2025-12-14
More xMath3 Cleanup

daft7:
More cleanup for this wonderful file. Some stuff was leftover from when I worked on it February, now tidied up along with some small progress on a few functions.


JoshSanch (approved):
Awesome work - this is exactly the kind of contribution I am excited to see as the decomp project approaches the end stages.

PR #713, opened 2025-12-21
Matched 10 small functions

AARJL:
Created namespace for parse_tag_pause, reset_tag_pause, and trigger_pause and matched them all

Also matched start_state_type::stop, start_state_type::update, next_state_type::stop, stop_state_type::start, stop_state_type::stop, stop_state_type::update, wait_context::reset_type

Defined the following structs

next_state_type
start_state_type
stop_state_type

Add variable reset_type to wait_context struct

### DIFF
--- a/src/SB/Game/zTalkBox.h
+++ b/src/SB/Game/zTalkBox.h
@@ -150,9 +150,10 @@ namespace
     struct state_type
     {
         state_enum type;
-
-        void start();
-        void stop();
+        
+        state_type(state_enum t);
+        virtual void start();
+        virtual void stop();
     };
 
     struct next_state_type
--- a/src/SB/Game/zTalkBox.h
+++ b/src/SB/Game/zTalkBox.h
@@ -155,6 +155,24 @@ namespace
         void stop();
     };
 
+    struct next_state_type
+    {
+        void stop();
+    };
+
+    struct start_state_type
+    {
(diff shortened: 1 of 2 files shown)


src/SB/Game/zTalkBox.h
```diff
@@ -150,9 +150,28 @@ namespace
...
+
+    struct next_state_type
+    {
+        void stop();
+    };
 
+    struct start_state_type
+    {
+        void stop();
+        signed char update(xScene&, float);
```
JoshSanch: Although these arguments aren't used and thus don't need to have parameter names for referencing inside the function, for these update functions I would like us to explicitly include parameter names, as they are the most likely to be updated by modders in the future and we've already seen what these arguments are named even if they were stripped from the DWARF data. 

Please update the signature to this:

```suggestion
        S8 update(xScene& scn, F32 dt);
```

JoshSanch (changes requested):
Nice work :)

Just some clean-up feedback - we have a types.h file which contains custom type definitions for primitive C/C++ types like unsigned char and signed long, etc. Please go through all of the functions and variables here and replace the primitive types with the shorter names from the types.h file. We use those so that it's easy to modify the types for different platforms as needed.

Thanks for contributing!

JoshSanch (approved):

PR #714, opened 2025-12-23
zCombo: Progress

bluisblu:
Seems equivalent to me, with some float/sda issues left and a switch statement's jump table needing to be matched, but leaving marked as NonMatching for now.

### DIFF
--- a/src/SB/Game/zCombo.cpp
+++ b/src/SB/Game/zCombo.cpp
@@ -35,7 +35,7 @@ static xVec3 sUnderCamPos;
 static ztextbox* sHideText[5];
 
 /* .sbss */
-widget_chunk* comboHUD;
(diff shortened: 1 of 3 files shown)


src/SB/Game/zCombo.cpp
```diff
@@ -35,7 +35,7 @@ static xVec3 sUnderCamPos;
 static ztextbox* sHideText[5];
 
 /* .sbss */
-static widget_chunk* comboHUD;
+widget_chunk* comboHUD;
```
JoshSanch: If this is global, it should also be extern'd inside of the zCombo.h header so that it is properly exported and accessible from other files.

JoshSanch (changes requested):

src/SB/Game/zCombo.cpp
bluisblu: That doesn't work for some reason, left a comment instead for now.

PR #716, opened 2026-01-04
zTalkBox Updates

AARJL:
- Made small cosmetic changes for empty space and neatness
- 
- Moved reset_auto_wait to lower down the file
- 
- Changed "unsigned long" to "size_t" for ztalkbox::load
- 
- ztalkbox::load_settings(xIniFile& ini) at 95%, seems to be missing a string but could not solve in scratch
- 
- read_bool to 99%, not sure how to fix the issue with the negative and positive arrays
- 
- parse_tag_trap to 99%, I could get to 100% when the if statements are triple nested and goto is used. Might be a better way. 
- 
- reset_tag_sound, reset_tag_trap, reset_tag_allow_quit, trigger_allow_quit, trigger_auto_wait all done at 100%
- 
- wait_context& wait_context::operator=(const wait_context& rhs)... yeah....
- 
- stop_audio_effect to 32%, still working on this
- 
- reset_auto_wait, reset_tag_auto_wait, and trigger_trap at 100%
- 
- Added various offsets in zTalkBox.h
- 
- Added struct sound_context in zTalkBox.h

### DIFF
--- a/src/SB/Game/zTalkBox.cpp
+++ b/src/SB/Game/zTalkBox.cpp
@@ -1,6 +1,7 @@
 #include "xstransvc.h"
 #include "zTalkBox.h"
-
+#include "xDebug.h"
+#include "zEntPlayer.h"
 #include <types.h>
 
 namespace
@@ -102,10 +103,6 @@ namespace
     }
 
     void flush_triggered()
-    {
-    }
-
-    void reset_auto_wait()
     {
     }
 
@@ -366,21 +363,115 @@ void ztalkbox::permit(U32 add_flags, U32 remove_flags)
     shared.permit |= add_flags;
 }
 
-void ztalkbox::load(xBase& data, xDynAsset& asset, unsigned long)
+void ztalkbox::load(xBase& data, xDynAsset& asset, size_t)
 {
     ((ztalkbox&)data).load((const ztalkbox::asset_type&)asset);
 }
 
+void ztalkbox::load_settings(xIniFile& ini) //TODO
+{
+    shared.volume = xIniGetFloat(&ini, "talk_box.volume", 2.0f);
+    xDebugAddTweak("Talk Box|\u{1}Globals|volume", &shared.volume, 0.1f, 10.0f, NULL, NULL, 0);
+}
+
 namespace
 {
+    static U8 read_bool(const substr& s, bool def) //TODO 99%
+    {
+        extern const substr negative[6];
+        extern const substr positive[6];
+        if (def)
+        {
+            for (U32 i = 0; i < 6; ++i)
+            {
+                if (icompare(s, negative[i]) == 0)
+                    return 0;
+            }
+            return 1;
+        }
+        else
+            for (U32 i = 0; i < 6; ++i)
+            {
+                if (icompare(s, positive[i]) == 0)
+                    return 1;
+            }
+        return 0;
+    }
     static void parse_tag_pause(xtextbox::jot&, const xtextbox&, const xtextbox&,
                                 const xtextbox::split_tag&)
     {
     }
 
-    static void reset_tag_pause(xtextbox::jot&, const xtextbox&, const xtextbox&,
-                                const xtextbox::split_tag&)
+    static void parse_tag_trap(xtextbox::jot& j, const xtextbox&, const xtextbox&,
+                               const xtextbox::split_tag& ti) //TODO 99%
     {
+        U8 c = 0;
+
+        if (ti.action.size == 1 && ti.action.text[0] == '=')
+        {
+            if (read_bool(ti.value, 1) != 0)
+            {
+                c = 1;
+            }
+        }
+
+        *(U8*)&j.context = c;
+    }
+
+    static void reset_tag_pause(xtextbox::jot&, const xtextbox&, const xtextbox&,
+                                const xtextbox::split_tag&) //DONE
+    {
+    }
+    static void reset_tag_sound(xtextbox::jot& j, const xtextbox&, const xtextbox& tb,
+                                const xtextbox::split_tag&) //DONE
+    {
+        if (!shared.active)
+        {
+            return;
+        }
+
+        if (&shared.active->dialog_box->tb != &tb)
+        {
+            return;
+        }
+
+        sound_context& c = *(sound_context*)j.context;
+
+        j.context_size = 24;
+
+        c.id = 0;
+        c.action = sound_context::ACTION_SET;
+    }
+    static void reset_tag_trap(xtextbox::jot& j, const xtextbox&, const xtextbox& ctb,
+                               const xtextbox::split_tag&) //DONE
+    {
+        if (!shared.active)
+        {
+            return;
+        }
+
+        if (&shared.active->dialog_box->tb != &ctb)
+        {
+            return;
+        }
+
+        *(bool*)&j.context = (shared.active->asset->trap != 0);
+    }
+
+    static void reset_tag_allow_quit(xtextbox::jot& j, const xtextbox&, const xtextbox& ctb,
+                                     const xtextbox::split_tag&) //DONE
+    {
+        if (!shared.active)
+        {
+            return;
+        }
+
+        if (&shared.active->dialog_box->tb != &ctb)
+        {
+            return;
+        }
+
+        *(bool*)&j.context = (shared.active->asset->allow_quit != 0);
     }
 
     U8 trigger_pause(const xtextbox::jot&)
@@ -388,9 +479,85 @@ namespace
         return 1;
     }
 
+    U8 trigger_allow_quit(const xtextbox::jot& j) //DONE
+    {
+        shared.allow_quit = (j.context != NULL);
+
+        return 1;
+    }
+    U8 trigger_auto_wait(const xtextbox::jot& j) //DONE
+    {
+        shared.auto_wait = *(const wait_context*)j.context;
+
+        return 1;
+    }
     state_type::state_type(state_enum t)
     {
         type = t;
+    }
+
+    wait_context& wait_context::operator=(const wait_context& rhs) //FIXME
+    {
+        type = rhs.type;
+        need = rhs.need;
+        delay = rhs.delay;
+        event_mask = rhs.event_mask;
+        query = rhs.query;
+        return *this;
+    }
+
+    void stop_audio_effect() //TODO
+    {
+        if (!shared.active)
+        {
+            return;
+        }
+
+        if (shared.active->asset == 0)
+        {
+            return;
+        }
+    }
+    static void reset_auto_wait() //DONE
+    {
+        const ztalkbox::asset_type* a = shared.active->asset;
+
+        shared.auto_wait.type.time = a->auto_wait.type.time;
+        shared.auto_wait.type.prompt = a->auto_wait.type.prompt;
+        shared.auto_wait.type.sound = a->auto_wait.type.sound;
+        shared.auto_wait.type.event = a->auto_wait.type.event;
+
+        shared.auto_wait.delay = a->auto_wait.delay;
+
+        shared.auto_wait.need = 0;
+
+        if (a->auto_wait.which_event <= 0 || a->auto_wait.which_event >= 32)
+        {
+            shared.auto_wait.event_mask = -1;
+        }
+        else
+        {
+            shared.auto_wait.event_mask = 1u << a->auto_wait.which_event;
+        }
+
+        shared.auto_wait.query = Q_SKIP;
+    }
+    static void reset_tag_auto_wait(xtextbox::jot& j, const xtextbox&, const xtextbox& ctb,
+                                    const xtextbox::split_tag&) //DONE
+    {
+        if (!shared.active)
+        {
+            return;
+        }
+        if (&shared.active->dialog_box->tb != &ctb)
+        {
+            return;
+        }
+
+        wait_context& c = *(wait_context*)j.context;
+        j.context_size = sizeof(wait_context);
+        reset_auto_wait();
+        c = shared.auto_wait;
     }
 
 } // namespace
@@ -424,4 +591,17 @@ S8 stop_state_type::update(xScene& scn, F32 dt)
 void wait_context::reset_type()
 {
     *(U16*)&this->type = 0;
+}
+
+static U8 trigger_trap(const xtextbox::jot& j) //DONE
+{
+    if (j.context != 0)
+    {
+        zEntPlayerControlOff(CONTROL_OWNER_TALK_BOX);
+    }
+    else
+    {
+        zEntPlayerControlOn(CONTROL_OWNER_TALK_BOX);
+    }
+    return 1;
 }
--- a/src/SB/Game/zTalkBox.cpp
+++ b/src/SB/Game/zTalkBox.cpp
@@ -373,18 +373,17 @@ void ztalkbox::load(xBase& data, xDynAsset& asset, unsigned long)
 
 namespace
 {
-
-    static void parse_tag_pause(xtextbox::jot&, const xtextbox&, const xtextbox&, const xtextbox::split_tag&)
+    static void parse_tag_pause(xtextbox::jot&, const xtextbox&, const xtextbox&,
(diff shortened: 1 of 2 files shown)


src/SB/Game/zTalkBox.cpp
```diff
@@ -366,33 +363,203 @@ void ztalkbox::permit(U32 add_flags, U32 remove_flags)
     shared.permit |= add_flags;
 }
 
-void ztalkbox::load(xBase& data, xDynAsset& asset, unsigned long)
+void ztalkbox::load(xBase& data, xDynAsset& asset, size_t)
```
JoshSanch: Is there a reason you made this change? It differs from the signature in objdiff which already matched:
`ztalkbox::load(xBase&, xDynAsset&, unsigned long)`

src/SB/Game/zTalkBox.cpp
```diff
@@ -366,33 +363,203 @@ void ztalkbox::permit(U32 add_flags, U32 remove_flags)
...
 {
     ((ztalkbox&)data).load((const ztalkbox::asset_type&)asset);
 }
 
+void ztalkbox::load_settings(xIniFile& ini) //TODO
+{
+    shared.volume = xIniGetFloat(&ini, "talk_box.volume", 2.0f);
+    xDebugAddTweak("Talk Box|\u{1}Globals|volume", &shared.volume, 0.1f, 10.0f, NULL, NULL, 0);
```
JoshSanch: Objdiff displays literal characters a little strange - we haven't figured out yet why these particular strings are this way but this is probably a better match for this one:
```suggestion
    xDebugAddTweak("Talk Box|\01Globals|volume", &shared.volume, 0.1f, 10.0f, NULL, NULL, 0);
```

src/SB/Game/zTalkBox.cpp
```diff
@@ -366,33 +363,203 @@ void ztalkbox::permit(U32 add_flags, U32 remove_flags)
     shared.permit |= add_flags;
 }
 
-void ztalkbox::load(xBase& data, xDynAsset& asset, unsigned long)
+void ztalkbox::load(xBase& data, xDynAsset& asset, size_t)
 {
     ((ztalkbox&)data).load((const ztalkbox::asset_type&)asset);
 }
 
+void ztalkbox::load_settings(xIniFile& ini) //TODO
```
JoshSanch: No need to leave these TODO comments - objdiff and our reporting tooling will show these as not matching 100% so it's redundant here.

src/SB/Game/zTalkBox.cpp
```diff
@@ -366,33 +363,203 @@ void ztalkbox::permit(U32 add_flags, U32 remove_flags)
...
+{
+    shared.volume = xIniGetFloat(&ini, "talk_box.volume", 2.0f);
+    xDebugAddTweak("Talk Box|\u{1}Globals|volume", &shared.volume, 0.1f, 10.0f, NULL, NULL, 0);
+}
+
 namespace
 {
+    static U8 read_bool(const substr& s, bool def) //TODO 99%
```
JoshSanch: Same comment as earlier here - no need for the TODO.

src/SB/Game/zTalkBox.cpp
```diff
@@ -366,33 +363,203 @@ void ztalkbox::permit(U32 add_flags, U32 remove_flags)
...
+    {
+        extern const substr negative[6];
+        extern const substr positive[6];
```
JoshSanch: These are definitely not extern - you can see in objdiff that there's data for these two arrays within the file. Please work with the Discord to figure these ones out.

Image

src/SB/Game/zTalkBox.cpp
```diff
@@ -366,33 +363,203 @@ void ztalkbox::permit(U32 add_flags, U32 remove_flags)
...
+                    return 0;
+            }
+            return 1;
+        }
+        else
+            for (U32 i = 0; i < 6; ++i)
+            {
+                if (icompare(s, positive[i]) == 0)
+                    return 1;
+            }
```
JoshSanch, line 398: Minor style comment to keep this in line with the rest of the project - add enclosing brackets instead of relying on the indentation here.
```suggestion
        else
        {
            for (U32 i = 0; i < 6; ++i)
            {
                if (icompare(s, positive[i]) == 0)
                    return 1;
            }
        }
```

src/SB/Game/zTalkBox.cpp
```diff
@@ -366,33 +363,203 @@ void ztalkbox::permit(U32 add_flags, U32 remove_flags)
...
+        return 0;
+    }
     static void parse_tag_pause(xtextbox::jot&, const xtextbox&, const xtextbox&,
                                 const xtextbox::split_tag&)
     {
     }
 
+    static void parse_tag_trap(xtextbox::jot& j, const xtextbox&, const xtextbox&,
```
JoshSanch, line 407: Please remove the TODO in line with the other review comments.

src/SB/Game/zTalkBox.cpp
```diff
@@ -366,33 +363,203 @@ void ztalkbox::permit(U32 add_flags, U32 remove_flags)
...
+                c = 1;
+            }
+        }
+
+        *(U8*)&j.context = c;
+    }
+
     static void reset_tag_pause(xtextbox::jot&, const xtextbox&, const xtextbox&,
-                                const xtextbox::split_tag&)
+                                const xtextbox::split_tag&) //DONE
```
JoshSanch: Please remove the DONE comment in line with the other review comments.

src/SB/Game/zTalkBox.cpp
```diff
@@ -366,33 +363,203 @@ void ztalkbox::permit(U32 add_flags, U32 remove_flags)
...
     {
     }
+    static void reset_tag_sound(xtextbox::jot& j, const xtextbox&, const xtextbox& tb,
+                                const xtextbox::split_tag&) //DONE
```
JoshSanch: Please remove the DONE comment in line with the other review comments.

src/SB/Game/zTalkBox.cpp
```diff
@@ -366,33 +363,203 @@ void ztalkbox::permit(U32 add_flags, U32 remove_flags)
...
+
+        sound_context& c = *(sound_context*)j.context;
+
+        j.context_size = 24;
+
+        c.id = 0;
+        c.action = sound_context::ACTION_SET;
+    }
+    static void reset_tag_trap(xtextbox::jot& j, const xtextbox&, const xtextbox& ctb,
+                               const xtextbox::split_tag&) //DONE
```
JoshSanch: Please remove the DONE comment in line with the other review comments.

src/SB/Game/zTalkBox.cpp
```diff
@@ -366,33 +363,203 @@ void ztalkbox::permit(U32 add_flags, U32 remove_flags)
...
+    }
     static void parse_tag_pause(xtextbox::jot&, const xtextbox&, const xtextbox&,
                                 const xtextbox::split_tag&)
     {
     }
 
+    static void parse_tag_trap(xtextbox::jot& j, const xtextbox&, const xtextbox&,
+                               const xtextbox::split_tag& ti) //TODO 99%
+    {
+        U8 c = 0;
```
JoshSanch: Given that read_bool is reading a boolean, it's more likely that `c` should have type `bool`. It would line up the `bool` casts in the subsequent functions. Let's use that instead and see if it matches.

src/SB/Game/zTalkBox.cpp
```diff
@@ -366,33 +363,203 @@ void ztalkbox::permit(U32 add_flags, U32 remove_flags)
...
+
+        *(bool*)&j.context = (shared.active->asset->allow_quit != 0);
+    }
 
     U8 trigger_pause(const xtextbox::jot&)
     {
         return 1;
     }
 
+    U8 trigger_allow_quit(const xtextbox::jot& j) //DONE
```
JoshSanch: Please remove the DONE comment in line with the other review comments.

src/SB/Game/zTalkBox.cpp
```diff
@@ -366,33 +363,203 @@ void ztalkbox::permit(U32 add_flags, U32 remove_flags)
...
+    {
+        shared.allow_quit = (j.context != NULL);
+
+        return 1;
+    }
+    U8 trigger_auto_wait(const xtextbox::jot& j) //DONE
```
JoshSanch: Please remove the DONE comment in line with the other review comments.

src/SB/Game/zTalkBox.cpp
```diff
@@ -366,33 +363,203 @@ void ztalkbox::permit(U32 add_flags, U32 remove_flags)
...
+        {
+            return;
+        }
+
+        if (shared.active->asset == 0)
+        {
+            return;
+        }
+    }
+    static void reset_auto_wait() //DONE
```
JoshSanch: Please remove the DONE comment in line with the other review comments.

src/SB/Game/zTalkBox.cpp
```diff
@@ -366,33 +363,203 @@ void ztalkbox::permit(U32 add_flags, U32 remove_flags)
...
+        }
+        else
+        {
+            shared.auto_wait.event_mask = 1u << a->auto_wait.which_event;
+        }
+
+        shared.auto_wait.query = Q_SKIP;
+    }
+    static void reset_tag_auto_wait(xtextbox::jot& j, const xtextbox&, const xtextbox& ctb,
+                                    const xtextbox::split_tag&) //DONE
```
JoshSanch: Please remove the DONE comment in line with the other review comments.

src/SB/Game/zTalkBox.cpp
```diff
@@ -425,3 +592,16 @@ void wait_context::reset_type()
 {
     *(U16*)&this->type = 0;
 }
+
+static U8 trigger_trap(const xtextbox::jot& j) //DONE
```
JoshSanch: Please remove the DONE comment in line with the other review comments.

src/SB/Game/zTalkBox.h
```diff
@@ -305,6 +306,43 @@ namespace
...
+
+    struct sound_context
+    {
+        // total size: 0x18
+        U32 id; // offset 0x0, size 0x4
+        enum
+        {
+            ACTION_SET = 0,
+            ACTION_PUSH = 1,
+            ACTION_POP = 2,
```
JoshSanch: When enums start at 0 and increment by one, we leave them implicitly defined to avoid visual noise. Opt for this instead, please:
```suggestion
            ACTION_SET,
            ACTION_PUSH,
            ACTION_POP,
```

src/SB/Game/zTalkBox.h
```diff
@@ -305,6 +306,43 @@ namespace
...
+        } action : 8; // offset 0x4, size 0x4
+        enum
+        {
+            TYPE_INVALID = 0,
+            TYPE_VOLUME = 1,
+            TYPE_TARGET = 2,
+            TYPE_ORIGIN = 3,
```
JoshSanch: Apply general enum styling feedback from the other one here as well, please

src/SB/Game/zTalkBox.h
```diff
@@ -305,6 +306,43 @@ namespace
...
+        } type : 8; // offset 0x4, size 0x4
+        enum
+        {
+            SOURCE_MEMORY = 0,
+            SOURCE_STREAM = 1,
```
JoshSanch: Apply general enum styling feedback from the other one here as well, please

JoshSanch (changes requested):
Nice progress :)

Left some feedback so we can work to get this in a merge-ready state.

src/SB/Game/zTalkBox.cpp
AARJL: I think I made this change because at some point I was asked to use the types.h file when defining variables as signed or unsigned. I guess this does not apply to function signatures.

src/SB/Game/zTalkBox.cpp
AARJL: I made the change but no change in objdiff.

src/SB/Game/zTalkBox.cpp
AARJL: I left these in by mistake, they were for me when I was coding to remember what I was doing. I will remove them along with the DONE comments.

src/SB/Game/zTalkBox.cpp
AARJL: See above.

src/SB/Game/zTalkBox.cpp
AARJL: See above.

src/SB/Game/zTalkBox.cpp
AARJL: See above.

src/SB/Game/zTalkBox.h
AARJL: For these I simply copy and pasted from the dwarf file. I will go ahead and make the changes for all enums from here.

src/SB/Game/zTalkBox.h
AARJL: See above.

src/SB/Game/zTalkBox.h
AARJL: See above.

src/SB/Game/zTalkBox.cpp
AARJL: Yeah I see this but was not sure what to make of it. I will reach out in Discord. Given that I can't seem to use decomp.me for this function due to namespacing issues, what is the best way to get help with that? Most of zTalkBox is in an anonymous namespace and that makes it difficult to reach out for help using that tool.

src/SB/Game/zTalkBox.cpp
AARJL: Got it. I must have overlooked that one.

src/SB/Game/zTalkBox.cpp
AARJL: I attempted using ``bool`` for this but no change. I tried using true and false along with 0 and 1, objdiff shows no difference.

AARJL:
Any comment on ``` wait_context& wait_context::operator=(const wait_context& rhs)``` ?

I am not sure why it didn't show up as a change here, but we briefly went over this during our VC and I was asked to include it in my next PR to go over it. Line 501

src/SB/Game/zTalkBox.cpp
JoshSanch: Might be a vc screenshare or just sharing pics as we go. I have some thoughts already since it'll have some form like:
```suggestion
        const substr positive[6] = {
            {"string1", 1},
            ...
        };
```

src/SB/Game/zTalkBox.cpp
JoshSanch: Ah I see - good instincts then, but we don't want to use `size_t` in this instance. Use lowercase `u32` instead. `size_t` has a specific semantic meaning wherein it is the size the particular CPU architecture uses for iterator variables in bits. Since this argument isn't used as an iterator for a loop or anything, `u32` is more appropriate.

This article is a solid discussion on `size_t` to reference: https://www.embedded.com/why-size_t-matters/

src/SB/Game/zTalkBox.cpp
JoshSanch: Let's prefer the changed code but share in the Discord somehow so we can help you finish the match. Might just be that the string data memory location is off but the actual code matches.

JoshSanch:
> Any comment on ` wait_context& wait_context::operator=(const wait_context& rhs)` ?
> 
> I am not sure why it didn't show up as a change here, but we briefly went over this during our VC and I was asked to include it in my next PR to go over it. Line 501

I'm going to spend some time seeing if I can get this to match on my own.

AARJL:
```read_bool``` went to 0% when trying to make changes to it, I'll work with the team on that.

JoshSanch:
Some small nits that could improve this but nothing that should prevent it from being merged for now. Going to merge.

PR #717, opened 2026-01-14
zNPCTypePrawn: matches on zNPCPrawn, aqua_beam, and helpers

aokennard:
Hi

### DIFF
--- a/src/SB/Game/zNPCTypePrawn.cpp
+++ b/src/SB/Game/zNPCTypePrawn.cpp
@@ -108,7 +108,7 @@ namespace
         F32 w;
         F32 h;
 
-        void create(S32, S32);
+        bool create(S32, S32);
         void destroy();
         void set_background(iColor_tag);
         void set_model_texture(xModelInstance&);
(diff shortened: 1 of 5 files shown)


SquareMan:
Hi! Welcome! It looks like you changed some of the existing APIs a bit. That's usually a big redflag but at a glance the changes actually seem plausible. Unfortunately it looks like there's some existing callsites that need to be fixed up preventing the build from succeeding and generating our regression reports. Once you fix that one of us will be happy to fully review this.

aokennard:
> Hi! Welcome! It looks like you changed some of the existing APIs a bit. That's usually a big redflag but at a glance the changes actually seem plausible. Unfortunately it looks like there's some existing callsites that need to be fixed up preventing the build from succeeding and generating our regression reports. Once you fix that one of us will be happy to fully review this.

Thanks! It turns out the actual implementation is a little weird, so I changed it to be accurate with what objdiff had - multiple push_front() usages, some with wrappers around the no-args implementation. Callsites in other files should be building now

src/SB/Game/zNPCTypePrawn.cpp
```diff
@@ -119,12 +119,109 @@ namespace
...
+            destroy();
+            return FALSE;
+        }
+
+        RpWorldAddCamera(world, cam);
+        _rwObjectHasFrameSetFrame(cam, RwFrameCreate());
+
+        xVec2 windowSize;
+        windowSize.x = 1.0;
+        windowSize.y = 1.0;
```
JoshSanch, line 160: Do these need to be floats instead of doubles?

JoshSanch (approved):
Changes look good! I approved but left one bit of feedback in case we can increase the matchingness a little bit.

src/SB/Game/zNPCTypePrawn.cpp
aokennard: They can be, but this doesn't improve the match%

PR #720, opened 2026-02-12
zNPCTypeCommon progress

bluisblu:
Still need some weak functions to show up, data should be 100%, and code is basically 99%


JoshSanch (approved):
Thanks for the contribution 🙂

PR #725, opened 2026-03-01
Major zTalkBox Progress

AARJL:
100% matched all but 13 functions in zTalkBox.cpp
Ensured DWARF local variables were used when possible
Reordered many functions to be aligned with objdiff where possible

Many of the smaller data sections were matched including StringBase and rodata as a whole
Made changes to zTalkBox.h along with some other files that zTalkBox.cpp relied on such as xEnt.h, containers.h, xFont.h, xSnd.h, and zTextBox.h


dwarf/SB/Game/zTalkBox.cpp
JoshSanch, line 1: This file shouldn't have any changes, seems like these are all formatting. To simplify reviewing this change, could you revert all the style changes made to this one? 

A command like this might be helpful (haven't tested it, run at your own risk):
`git restore --source=$(git rev-list -2 HEAD -- dwarf/SB/Game/zTalkBox | tail -n1) dwarf/SB/Game/zTalkBox`

src/SB/Core/x/containers.h
```diff
@@ -245,8 +250,7 @@ template <class T, U32 N> struct fixed_queue
 
         iterator* operator+=(S32 value)
         {
-            value += _it;
-            _it = (value + N) & N;
+            _it = (_it + value) % (N + 1);
```
JoshSanch: Seems like this cause a regression with another function
`fixed_queue<aqua_beam::ring_segment, 31>::iterator::operator+=(int)`.

Please revert or see if there's a version that will 100% match both.

src/SB/Core/x/xSnd.h
```diff
@@ -95,7 +95,75 @@ template <S32 N> struct sound_queue
...
+            {
+                return false;
+            }
+        }
+        return true;
+    }
+    else
+    {
+        S32 i = 0;
+        for (; i < index; i++)
```
JoshSanch: If these don't match differently by declaring i earlier and assigning zero as part of the loop statement, that would be preferred here.

JoshSanch (changes requested):
Lots of progress with this one - left some initial comments while I spend some time reviewing zTalkBox.cpp.

Thanks for the submission!

src/SB/Core/x/containers.h
AARJL: Found a version that 100% matches both, will sync with this PR for review.

dwarf/SB/Game/zTalkBox.cpp
AARJL: Successfully restored DWARF back to it's original state. Sending the commit

src/SB/Core/x/xSnd.h
AARJL: Successfully resolved this, committing now.

src/SB/Game/zTalkBox.cpp
```diff
@@ -476,134 +704,1306 @@ namespace
...
+
+        if (mask == 0)
+        {
+            mask = 0xFFFFFFFF;
+        }
+
+        e.stop_wait(mask);
+    }
+
+    S32 cb_dispatch(xBase*, xBase* to, U32 event, const F32* argf, xBase*)
```
JoshSanch, line 1159: Please update the numbers in here to align with their `en_xEventTags` enumeration values, found in `xEvent.h`.

src/SB/Game/zTalkBox.cpp
```diff
@@ -476,134 +704,1306 @@ namespace
...
+
+            xBase* mychild = zSceneFindObject(link->dstAssetID);
+            if (mychild)
+            {
+                MasterLoveSlave(mychild, isBeginning);
+            }
+        }
+    }
+}
+void ztalkbox::MasterLoveSlave(xBase* slave, S32 starting)
```
JoshSanch, line 1549: Please update the cases in here to use the corresponding enumeration values from the `en_ZBASETYPE` enum in `zBase.h`.

JoshSanch (changes requested):

src/SB/Core/x/xSnd.h
```diff
@@ -95,7 +95,74 @@ template <S32 N> struct sound_queue
     void play(U32 id, F32 vol, F32 pitch, U32 priority, U32 flags, U32 parentID,
               sound_category snd_category);
     void push(U32 id);
+    void clear();
+    S32 size() const;
+    void pop();
+    bool playing(S32 index, bool streaming) const;
+    U32 recent(S32 index) const;
```
JoshSanch: The following implementations can be moved into this class so that we can avoid having to declare template functions over and over, and avoid the redundancy of declaring them here and also declaring them with their implementation.

JoshSanch (changes requested):

src/SB/Game/zTalkBox.cpp
AARJL: Updated the case values, committing soon

src/SB/Core/x/xSnd.h
AARJL: Moved their implementations to withing the class and deleted the declarations. Committing now

src/SB/Game/zTalkBox.cpp
AARJL: Updated this and found the rest of the match, at 100% now!

PR #738, opened 2026-04-23
zCinematic small matches

LSLees:


src/SB/Game/zNPCFXCinematic.cpp
```diff
@@ -447,6 +445,24 @@ void NCINBeNosey::CanRenderNow()
...
 
+void zNPCFXShutdown()
+{
+}
+
+void NCINBeNosey::Init(const zCutsceneMgr* m, NCINEntry* e, S32 i)
+{
+    this->use_fxtab = e;
+    this->x = i;
+    this->use_csnmgr = (zCutsceneMgr*)m;
```
SquareMan: Have you checked if the `use_csnmgr` field on this class can be `const zCutsceneMgr*` here instead of `zCutsceneMgr*` to avoid the need to cast away constness here?

src/SB/Game/zNPCFXCinematic.h
```diff
@@ -202,9 +202,12 @@ struct NCINBeNosey : XCSNNosey
 {
     zCutsceneMgr* use_csnmgr;
     NCINEntry* use_fxtab;
+    S32 x;
```
SquareMan: I don't see this field in the dwarf for this class. Let's review the asm that's accessing this field and see if we can figure out what's really going on.

SquareMan (changes requested):

PR #742, opened 2026-05-21
zNPCTypeVillager progress

bluisblu:
Data still needs work, .sdata2 is out of order and .rodata is missing zeroed out entries (possibly from deadstripped functions?)


JoshSanch:
🐐

PR #746, opened 2026-06-07
simple shadow :)

spongeis5:
PLEASE oh PLEASE review my code i dont wanna screw anything up 😔

### DIFF
--- a/src/SB/Core/x/xGlobals.h
+++ b/src/SB/Core/x/xGlobals.h
@@ -55,6 +55,10 @@ struct xGlobals
     U32 minVSyncCnt;
     U8 dontShowPadMessageDuringLoadingOrCutScene;
     U8 autoSaveFeature;
+    
+    U32 pad_0x6DC[1593];
+
+    xScene* sceneCur; 
 };
 
 #endif
--- a/src/SB/Core/x/xGlobals.h
+++ b/src/SB/Core/x/xGlobals.h
@@ -55,10 +55,6 @@ struct xGlobals
     U32 minVSyncCnt;
     U8 dontShowPadMessageDuringLoadingOrCutScene;
     U8 autoSaveFeature;
-    
-    U32 pad_0x6DC[1593];
-
-    xScene* sceneCur; 
 };
 
 #endif
--- a/src/SB/Core/x/xScene.h
+++ b/src/SB/Core/x/xScene.h
@@ -7,6 +7,8 @@
 #include "xMemMgr.h"
 #include "xRay3.h"
 
+struct xClumpCollBSPTree; 
+
 typedef const char* (*xSceneBase2NameCallback)(xBase*);
 typedef const char* (*xSceneID2NameCallback)(U32);
 typedef xBase* (*xSceneResolvIDCallback)(U32);
@@ -38,7 +40,13 @@ struct xScene
     xEnt** npcs;
     xEnt** act_ents;
     xEnt** nact_ents;
-    xEnv* env;
+    
+    // Offset: 0x44
+    union {
+        xEnv* env;
+        xClumpCollBSPTree* scenecoll;
+    };
+    
     xMemPool mempool;
     xSceneResolvIDCallback resolvID;
     xSceneBase2NameCallback base2Name;
--- a/src/SB/Core/x/xShadowSimple.cpp
+++ b/src/SB/Core/x/xShadowSimple.cpp
@@ -96,11 +96,13 @@ static int shadowRayEntCB(xEnt* ent, void* data)
                                   ((shadowRayEntData*)data)->line->start.y, 
                                   ((shadowRayEntData*)data)->line->start.z);
 
-                float dy = ((shadowRayEntData*)data)->line->end.y - start.y;
-                float t = (ent->bound.box.box.upper.y - start.y) / dy;
+                float end_y = ((shadowRayEntData*)data)->line->end.y;
+                float start_y = start.y;
+                float dy = end_y - start_y;
+                float t = (ent->bound.box.box.upper.y - start_y) / dy;
 
                 if (((shadowRayEntData*)data)->cache->shadowHeight > t && 
-                    ent->bound.box.box.upper.y > ((shadowRayEntData*)data)->line->end.y) {
+                    ent->bound.box.box.upper.y > end_y) {
                     
                     if (start.x > ent->bound.box.box.lower.x && start.x < ent->bound.box.box.upper.x &&
                         start.z > ent->bound.box.box.lower.z && start.z < ent->bound.box.box.upper.z) {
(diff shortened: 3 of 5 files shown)


src/SB/Core/x/xGlobals.h
```diff
@@ -55,6 +55,10 @@ struct xGlobals
     U32 minVSyncCnt;
     U8 dontShowPadMessageDuringLoadingOrCutScene;
     U8 autoSaveFeature;
+    
+    U32 pad_0x6DC[1593];
+
+    xScene* sceneCur;
```
SquareMan: This struct should not be modified, it is 99.99% likely to be fully correct. From memory I believe this field is actually in `zGlobals`, which embeds this struct as it's first member. This is probably also what's failing the build btw.

src/SB/Core/x/xShadowSimple.h
```diff
@@ -45,17 +55,20 @@ struct xShadowSimpleCache
...
+    U32 modelID;
+    U32 assetID;
+    U32 flags;
+};
+
 struct zSimpleShadowTableHeader {
-    // total size: 0x4
-    U32 num;
+    S32 num;
+    zSimpleShadowTableEntry entry[1];
```
SquareMan: There is no reason to use an array of size one here, just `zSimpleShadowTableEntry entry` is the same thing. This is also likely an incorrect change. Our struct defs primarily come from ps2 dwarf and are very accurate the non-platform specific code.

src/SB/Core/x/xShadowSimple.cpp
```diff
@@ -7,158 +7,683 @@
...
+#include "xGlobals.h"
+#include "xGrid.h"
+#include "xScene.h"
 
 #include "rpcollis.h"
 #include "xMath3.h"
 #include "xQuickCull.h"
+#include "iModel.h"
+
+extern xGlobals globals;
```
SquareMan: As touched on earlier, this should be a `zGlobals` not an `xGlobals`. If you simply include zGlobals.h you can remove this line as the header contains the correct extern definition.

SquareMan (changes requested):
Thanks for this and welcome to the project! I haven't fully reviewed the function code yet as there are some major structural issues introduced by these struct edits you've made. Feel free to ask for more help in the discord if you need it but once this PR passes CI one of us will give it a more thorough review.

spongeis5:
thank you so much for the feedback !!

src/SB/Core/x/xScene.h
```diff
@@ -38,7 +40,13 @@ struct xScene
     xEnt** npcs;
     xEnt** act_ents;
     xEnt** nact_ents;
-    xEnv* env;
+    
+    // Offset: 0x44
+    union {
+        xEnv* env;
+        xClumpCollBSPTree* scenecoll;
+    };
```
SquareMan, line 48: ok so this is pretty sus. We're gonna focus on figuring out what's really going on

src/SB/Core/x/xShadowSimple.cpp
```diff
@@ -7,158 +7,682 @@
...
+#include "iModel.h"
+
+extern xGrid colls_grid;
+extern xGrid colls_oso_grid;
+extern xGrid npcs_grid;
+
+struct shadowRayEntData {
+    xShadowSimpleCache* cache;
+    const RwLine* line;
+};
```
SquareMan, line 26: can you tell me where this struct came from?

src/SB/Core/x/xShadowSimple.cpp
```diff
@@ -7,158 +7,682 @@
...
+
+class xNPCBasic : public xEnt {
+public:
+    u32 SelfType() const;
+};
```
SquareMan, line 31: Missed this before. `xNPCBasic` is decarled in `xNPCBasic.h`, you should remove this and include that header instead.

src/SB/Core/x/xShadowSimple.cpp
```diff
@@ -7,158 +7,682 @@
...
+        xVec3Normalize(&xformnorm, &xformnorm);
+        norm = &xformnorm;
+    }
+    else
+    {
+        testdist = &cache->envHeight;
+        norm = (xVec3*)&tri->normal;
+    }
+
+    if (dist >= *testdist || (norm->y < 0.0871557f))
```
SquareMan, line 62: can you try using `DEG2RAD(5)` here instead of `0.0871557`?

src/SB/Core/x/xShadowSimple.cpp
```diff
@@ -7,158 +7,682 @@
...
+    isx.t.line.end.y = py - depth;
+    isx.t.line.end.z = pos->z;
+
+    RpClump* clump = *(RpClump**)clumpcoll;
+    
+    if (*(RpAtomic**)((char*)clump + 0x10)) {
+        xClumpColl_ForAllIntersections(*(xClumpCollBSPTree**)((char*)*(RpAtomic**)((char*)clump + 0x10) + 0x10), &isx, shadowRayCB, cache);
+    } else {
+        RpCollisionWorldForAllIntersections(*(RpWorld**)clump, &isx, shadowRayCB, cache);
+    }
```
SquareMan, line 206: This is a mess and needs to be cleaned. In doing so I think we can solve our scenecoll mystery with xScene. It's not unheard of to see crazy pointer math like this with Renderware, but usually it's handled by a macro from one of the rw headers if necessary.

src/SB/Core/x/xShadowSimple.cpp
```diff
@@ -7,158 +7,682 @@
...
+        sShadRasters[sShadVertCount / 6] = (RwRaster*)cache->raster;
+        
+        sShadVertCount = *(volatile u32*)&sShadVertCount + 6;
+    }
+}
+
+#pragma push
+#pragma optimization_level 1
+#pragma optimize_for_size on
+
```
SquareMan, line 388: This is probably not accurate. I'd rather it be removed at the cost of a lower match percentage but others might disagree.

src/SB/Core/x/xShadowSimple.cpp
```diff
@@ -7,158 +7,682 @@
...
         {
-            flags = sShadRaster->width;
+            flags = (U32)sShadRaster;
         }
-        cache->corner[1].x = flags;
-        cache->flags |= (S16)var_r20;
+
+        cache->raster = flags;
+        cache->flags |= (U16)var_r20;
         ent->model->shadowID = flags;
```
SquareMan, line 500: I think you should give this one another go. The last three lines here are a big tell. Looks like the current `flags` var is actually `raster`, `var_r20` is `flags`. Not sure about the last line since it seems to match but it's odd too.

src/SB/Core/x/xShadowSimple.h
```diff
@@ -45,17 +55,19 @@ struct xShadowSimpleCache
     xVec3 corner[4];
 };
 
+struct zSimpleShadowTableEntry {
+    U32 modelID;
+    U32 assetID;
+    U32 flags;
+};
+
```
SquareMan, line 63: Also curious about this struct, since it's not in DWARF.

SquareMan (changes requested):

src/SB/Core/x/xShadowSimple.cpp
ThePlayerRolo: > This is a mess and needs to be cleaned. In doing so I think we can solve our scenecoll mystery with xScene. It's not unheard of to see crazy pointer math like this with Renderware, but usually it's handled by a macro from one of the rw headers if necessary.

Thats like Ghidra level pointer math

PR #748, opened 2026-07-01
iSystem 99%

bluisblu:
The time getter functions aren't weak surprisingly, the definitions are genuinely in iSystem, but it still feels right to leave the declarations in iTime's header.


src/SB/Core/gc/iSystem.h
```diff
@@ -9,6 +10,24 @@ void* malloc(U32 __size);
...
+    U32 bufferPos; // 0x2844
+    S32 accessMode; // 0x2848
+};
+
+struct dlFSUnkGlobals // entirely fabricated
+{
+    RwFileFunctions fileFuncs; // 0x00
+    RwFreeList* unk_2C; // 0x2C
+};
+
```
JoshSanch: Please change these to use memory offset naming unless you know for a fact how a field is used.

JoshSanch (changes requested):
Looks good except for the changes mentioned. Great work!

src/SB/Core/gc/iSystem.h
bluisblu: Done 👍

PR #753, opened 2026-07-19
Replace pull_request_trigger

SquareMan:
Prevent pwn request

Looks like `pull_request_target` was used before because build-previous would fail with `pull_request`. Root cause was actually because build-previous wants to checkout main but it isn't fetched for pull requests when using the `pull_request` trigger. Fix is to force the checkout action to always checkout main.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,7 +2,7 @@ name: Build
 
 on:
   push:
-  pull_request_target:
+  pull_request:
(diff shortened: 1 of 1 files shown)


JoshSanch:
LGTM.. not sure if things will run for external contributor PRs but we can fail back quickly and safely so that's alright.

JoshSanch (approved):

PR #757, opened 2026-07-19
xCM progress

bluisblu:


src/SB/Core/x/xCM.cpp
JoshSanch, line 96: Is this meant to be a while loop?

src/SB/Core/x/xCM.cpp
```diff
@@ -73,12 +100,196 @@ static void xCMprep(xCreditsData* data)
...
+                    }
+                    break;
+                }
+                default:
+                    break;
+                }
+            }
+        }
+        cp = (xCMcredits*)hp;
+    }
```
JoshSanch, line 280: Could also be a while loop maybe

JoshSanch (commented):
Some nits, good stuff

JoshSanch (approved):

PR #763, opened 2026-07-25
New .clang-format files/header cleanup

bluisblu:


SquareMan:
What is the rationale for having separate clang-formats for gc/x/game code?

PR #764, opened 2026-07-25
zFX: zGooRenderAtomic

SquareMan:


src/SB/Game/zFX.cpp
```diff
@@ -435,12 +438,157 @@ void zFXGooUpdateInstance(zFXGooInstance* goo, F32 dt)
...
         }
     }
 }
 
+void zFXGooUpdate(F32 dt)
+{
+    int i;
+    zFXGooInstance* pGoo = &zFXGooInstances[0];
+
+    for (i = 0; i < 0x18; i++)
```
JoshSanch: Opt for decimal to be consistent with later constant usage

JoshSanch (changes requested):

src/SB/Game/zFX.cpp
SquareMan: ah true true true

JoshSanch (approved):

PR #767, opened 2026-07-28
zNPCTypeKingJelly: Additional Function and Data Matching

JoshSanch:


src/SB/Game/zNPCTypeKingJelly.cpp
```diff
@@ -195,6 +162,287 @@ namespace
...
+                sin_lookup[i] = isin(angle);
+                cos_lookup[i] = icos(angle);
+
+                const F32 angle_step = PI / 32.0f;
+                angle += angle_step;
+            }
+        }
+
+        S32 size = 0.5f + ((2.0f * radius * PI) / segment_length);
+        size = (size + 7) & ~7;
```
SquareMan, line 368: Does `ALIGN_NEXT` from macros.h work here?

SquareMan:
Have you checked the regressions? I think your branch just needs to be rebased

src/SB/Game/zNPCTypeKingJelly.cpp
JoshSanch: it does not :(

PR #776, opened 2026-08-07
A tiny bit of Plankton work

mattbruv:
Did some plankton work the other day, just gonna get it checked in because I may not do decomping work for a bit.

### DIFF
--- a/src/SB/Game/zNPCTypeBossPlankton.cpp
+++ b/src/SB/Game/zNPCTypeBossPlankton.cpp
@@ -1147,8 +1147,7 @@ void zNPCBPlankton::Process(xScene* xscn, F32 dt)
     update_turn(dt);
     update_move(dt);
     update_animation(dt); //uvar5 = update anim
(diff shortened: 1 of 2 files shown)


src/SB/Game/zNPCTypeBossPlankton.cpp
```diff
@@ -1147,8 +1148,7 @@ void zNPCBPlankton::Process(xScene* xscn, F32 dt)
     update_turn(dt);
     update_move(dt);
     update_animation(dt); //uvar5 = update anim
-    check_player_damage(); //uvar1 = check_player_damage
-    if (psy_instinct != 0) //psy_instinct isnt right, needs (uvar1 & 0xff)
+    if ((U8)check_player_damage())
```
SquareMan, line 1151: Did you consider changing this function's signature to return U8/char/bool or something or does that not work?

SquareMan (approved):

